PR #1, opened 2019-02-27
[cli tools] bulk of implementation

janedegtiareva:


unencrypted_file_system_keystore.js
```diff
@@ -0,0 +1,64 @@
+const fs = require('fs');
+const keyDir = './neardev';
+const keyFile = 'devkey.json';
```
janedegtiareva: this probably belongs in nearlib. I'll move it once stuff is more stable.

unencrypted_file_system_keystore.js
```diff
@@ -0,0 +1,64 @@
...
+const KeyPair = require('nearlib/signing/key_pair')
+
+/**
+ * Unencrypted file system key store.
+ */
+class UnencryptedFileSystemKeyStore {
+    constructor() {}
+
+    async setKey(accountId, keypair) {
+        if (!fs.existsSync(keyDir)){
```
vgrichina: it's better to avoid using `Sync` variant of fs functions, except for stuff which is only ever executed during process start up.

PR #2, opened 2019-03-01
[near-cli] fix deploy to work with latest devnet (param was removed)

janedegtiareva:


gulpfile.js
```diff
@@ -63,7 +63,8 @@ gulp.task('createDevAccount', async function(argv) {
 
 
 async function deployContractAndWaitForTransaction(accountId, contractName, data, near) {
-    const deployContractResult = await near.deployContract(accountId, contractName, data);
+    console.log(data[data.length-1]);
```
vgrichina: is this console.log still needed?

gulpfile.js
```diff
@@ -92,7 +93,8 @@ gulp.task('deploy', async function(argv) {
     };
 
     const near = await neardev.connect(options);
-    const contractData = [...fs.readFileSync('./out/main.wasm')];
+    console.log("deploying file ../tests/hello.wasm");
```
vgrichina: should use actual file name

PR #3, opened 2019-03-04
[nearshell] build works on standard files

janedegtiareva:


gulpfile.js
```diff
@@ -75,6 +86,17 @@ async function deployContractAndWaitForTransaction(accountId, contractName, data
     return waitResult;
 }
 
+// converts file.ts to file.near.ts
+function generateNearFileName(fileName) {
```
vgrichina: `fileName.replace(/.ts$/, '.near.ts')`

vgrichina: or if you want to be pedantic about path handling, it's in node's std: https://nodejs.org/api/path.html#path_path_basename_path_ext

PR #12, opened 2019-03-23
Depend on AS compiler remove-result branch

evgenykuzyakov:
Related to https://github.com/nearprotocol/assemblyscript/pull/26


evgenykuzyakov:
It's probably not needed as long as master branch near-shell depends on the master branch of assemblyscript

Issue #21, opened 2019-03-28
Support configuration of different deployment environments


vgrichina:
As a developer I don't want to specify full arguments to `near deploy` each time I need to deploy to different environment.

What I want to do instead is to have config file with necessary parameters for different env, e.g. like here:
https://github.com/sequelize/express-example/blob/master/config/config.js

As result deploy to production can be as easy as:

`NODE_ENV=production npm run deploy`

PR #22, opened 2019-04-01
small tweaks to settings.js which is used from main.html

janedegtiareva:
to make local ui development experience better


vgrichina:
I think it makes sense to align this with changes I've been making to make GH pages work well for:
https://github.com/nearprotocol/near-chess
https://github.com/nearprotocol/near-shell/commits/support-gh-pages

I think we need to agree on how to solve https://github.com/nearprotocol/near-shell/issues/21, cause otherwise it's very inconvenient to switch between local / shared devnet.

janedegtiareva:
> I think it makes sense to align this with changes I've been making to make GH pages work well for:
> https://github.com/nearprotocol/near-chess
> https://github.com/nearprotocol/near-shell/commits/support-gh-pages
> 
> I think we need to agree on how to solve #21, cause otherwise it's very inconvenient to switch between local / shared devnet.

re GH pages - settings.js looks pretty much identical except for setting the fiddle cookie. Can we actually change that? I don't want the real app code to say "fiddle" anywhere. Why can't we just read the values from the settings.js directly and bypass fiddle config? 

#21 - switching between envs is just passing node url at this point. I do not think that fixing this is higher priority than key management, so will not work on it right now.

vgrichina:
> Why can't we just read the values from the settings.js directly and bypass fiddle config?

I agree we shouldn't use cookie, that's just hack to make stuff work quickly. 

I'm thinking that it makes sense to change it as part of doing https://github.com/nearprotocol/near-shell/issues/21, so that e.g. we have few .json config files (one per environment).

janedegtiareva:
> > Why can't we just read the values from the settings.js directly and bypass fiddle config?
> 
> I agree we shouldn't use cookie, that's just hack to make stuff work quickly.
> 
> I'm thinking that it makes sense to change it as part of doing #21, so that e.g. we have few .json config files (one per environment).

Cool! This MR is going in that direction by adding 2 settings files. You still need to specify which one to use manually, but this is something that can be iterated on.

blank_project/src/index.html
```diff
@@ -7,11 +7,11 @@
 </head>
 <body style="background: #fff">
   <div class="container">
-    Contract says:
+    Contract says 2:
```
vgrichina: @janedegtiareva hmm, seems like typo

janedegtiareva: good catch. I use these sometimes to make sure i am looking at the latest stuff and forget to undo lol

PR #26, opened 2019-04-11
Resolve near library mappings in readFile

vgrichina:


vgrichina:
Merging this to unbreak `nearcore` build

PR #27, opened 2019-04-12
Remove dev keys from blank project

vgrichina:


janedegtiareva:
Could you also add the default dir to gitignore so that it doesn't accidentally get checked in again?

PR #28, opened 2019-04-12
Return error code / wire commands directly

vgrichina:
Resolve https://github.com/nearprotocol/near-shell/issues/23
Resolve https://github.com/nearprotocol/near-shell/issues/4

### DIFF
--- a/bin/near
+++ b/bin/near
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 const yargs = require('yargs');
-const main = require('.');
+const main = require('../');
 
 const exitOnError = async function(promise) {
   try {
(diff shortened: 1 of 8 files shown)


bin/near
```diff
@@ -1,9 +1,16 @@
 #!/usr/bin/env node
 
-const gulp = require('gulp');
```
janedegtiareva: Can we still do command line near .... after near -> bin/near?

vgrichina: @janedegtiareva do you mean whether it is in `PATH` when installed by `npm`? It's specified by `bin` config in `package.json`, I changed path there as well.

PR #32, opened 2019-04-24
Support config with settings for different NODE_ENV environments

vgrichina:
This is start of the work, only supports configs for `npm run deploy` / `npm start` and manual `near` usage.


bin/near
```diff
@@ -84,20 +65,44 @@ const newProject = {
...
 yargs // eslint-disable-line
+  .option('nodeUrl', {
+    desc: 'NEAR node URL',
+    type: 'string',
+    default: 'http://localhost:3030'
+  })
+  .option('helperUrl', {
+    desc: 'NEAR contract helper URL',
+    type: 'string',
+    default: 'http://localhost:3030'
```
evgenykuzyakov: should this be :3000?

vgrichina: should actually just be `null` I think (node should be used directly locally)

PR #35, opened 2019-04-29
Use config in test environment

vgrichina:


bin/near
```diff
@@ -78,13 +78,7 @@ const newProject = {
   handler: (argv) => exitOnError(main.newProject(argv))
 };
 
-const configPath = process.cwd() + '/src/config';
```
janedegtiareva: does this setup allow you to easily have different config for test and staging/prod?

vgrichina: Yes, it's switched based on `NODE_ENV` value https://github.com/nearprotocol/near-shell/pull/32/files#diff-f6be52936ac73f7f253f2ac186f3392aR5

PR #47, opened 2019-06-04
Add positional arguments to create_account and new_project

potatodepaulo:
Resolves https://github.com/nearprotocol/near-shell/issues/46


get-config.js
```diff
@@ -1,6 +1,6 @@
 
 module.exports = function getConfig() {
-    const configPath = process.cwd() + '/src/config';
+    const configPath = process.cwd() + '/src/config.js';
```
vgrichina: hmm, why did you have to change this?

index.js
```diff
@@ -35,18 +35,18 @@ exports.clean = async function() {
 // Only works for dev environments
 exports.createDevAccount = async function(options) {
     const keyPair = await KeyPair.fromRandomSeed();
-
+    let accountId = options.accountId;
```
vgrichina: nit: `const { accountId } = options;` (same tricks can be used to extract other options)

test/new_project.sh
```diff
@@ -1,10 +1,22 @@
...
-../bin/near new_project
+../../bin/near new_project
 npm install
 npm uninstall near-shell
-npm install ../
+npm install ../../
 NODE_ENV=development npm run test
+cd ..
+rm  -rf tmp-project
+# test generating new project in new dir
```
vgrichina: full test cycle is pretty slow, I think it's enough to test option with explicit directory full cycle and just check that both options write out `package.json`

index.js
potatodepaulo: Ah yeah, good call. Will add.

test/new_project.sh
potatodepaulo: Can run tests in the first cycle and will just check the package.json moves in the second.

get-config.js
potatodepaulo: Wasn't finding the config.js file locally. Worked when I changed it. 🤷‍♀

test/new_project.sh
```diff
@@ -16,7 +20,9 @@ rm  -rf tmp-project
...
-npm install
-npm uninstall near-shell
-npm install ../../
-NODE_ENV=development npm run test
\ No newline at end of file
+FILE=package.json
+if test -f "$FILE"; then
+  echo "${GREEN}$FILE exists. Have a cookie!"
+else
+  echo "${RED}ERROR: file not found."
```
vgrichina: need to add `exit 1` or it won't return error code (and so won't fail CI)

potatodepaulo: done

PR #51, opened 2019-06-06
Reenable strict mode so that shell gives helpful messages for unknown…

janedegtiareva:
… commands

### DIFF
--- a/bin/near
+++ b/bin/near
@@ -109,6 +109,7 @@ const newProject = {
 
 let config = require('../get-config')();
 yargs // eslint-disable-line
+  .strict()
   .option('nodeUrl', {
     desc: 'NEAR node URL',
     type: 'string',


janedegtiareva:
fix for https://github.com/nearprotocol/near-shell/issues/50, which is a regression

vgrichina:
AFAIK, this will break `near-shell` for cases when config is present, but given command doesn't accept all of the options in config.

ilblackdragon:
@janedegtiareva  can we define `strict` only on the top commands? e.g. `near blah` fails but not `near send --blah=whatever`?

janedegtiareva:
We can add all the config options to shell and expect no unrecognized config options. Opinions?

mikedotexe:
> We can add all the config options to shell and expect no unrecognized config options. Opinions?

This is essentially the approach I took here:
https://github.com/nearprotocol/near-shell/pull/262

Plus some other fun snags to work through.

PR #54, opened 2019-06-08
Nightshade update

ilblackdragon:


bin/near
```diff
@@ -80,7 +80,19 @@ const createAccount = {
       type: 'string',
       required: true
     }),
-  handler: (argv) => exitOnError(main.createDevAccount(argv))
+  handler: (argv) => exitOnError(main.createAccount(argv))
+};
+
+const viewAccount = {
```
vgrichina: https://github.com/nearprotocol/near-shell/commit/bf53541192855b03e71853dc00c4b6d5f28163ef

ilblackdragon: Btw, I'm fine with `view_account` instead of `state` command. You tell me what do you think is better.

vgrichina: `state` is shorter and makes sense

PR #58, opened 2019-06-17
Update AssemblyScript syntax

bowenwang1996:
Update assemblyscript syntax after the runtime change. Requires https://github.com/nearprotocol/assemblyscript/pull/36, https://github.com/nearprotocol/near-runtime-ts/pull/27, and https://github.com/nearprotocol/assemblyscript-json/pull/13.

### DIFF
--- a/blank_project/assembly/main.ts
+++ b/blank_project/assembly/main.ts
@@ -9,7 +9,7 @@ import { Greeter } from "./model.near";
 // using `export` keyword.
 
 export function hello(): string {
-  let greeter = new Greeter("hello");
+  let greeter = new Greeter("Hello");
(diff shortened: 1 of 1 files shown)


blank_project/assembly/main.ts
```diff
@@ -12,7 +9,7 @@ import { Greeter } from "./model.near";
 // using `export` keyword.
 
 export function hello(): string {
-  let greeter = new Greeter("Hello");
+  let greeter = new Greeter("hello");
```
vgrichina: why is this changed?

bowenwang1996: Not sure. I'll change it back

PR #61, opened 2019-07-02
Login with wallet though shell: command format

janedegtiareva:


index.js
```diff
@@ -105,3 +105,14 @@ exports.stake = async function(options) {
...
 }
+
+exports.login = async function(options) {
+    if (!options.walletUrl) {
+        console.log("Log in is not needed on this environment. Please use appropriate master account for shell operations.")
+    } else {
+        const title = 'Near Shell';
+        const keyPair = await KeyPair.fromRandom('ed25519');
+        console.log(`Please navigate to this url and follow the insturctions to log in: ` +
+            `${options.walletUrl}/login/?title=${title}&public_key=${keyPair.getPublicKey()}&account_id=${options.accountId}`);
```
vgrichina: nit: `title` generally needs to be URL-encoded. Prob best to use https://nodejs.org/api/querystring.html

bin/near
```diff
@@ -83,6 +83,23 @@ const createAccount = {
...
+const login = {
+  command: 'login <accountId>',
+  desc: 'create a developer account',
+  builder: (yargs) => yargs
+    .option('walletUrl', {
+      desc: 'Wallet URL to use for login',
+      type: 'string',
+      default: 'https://wallet.nearprotocol.com',
+    })
+    .option('accountId', {
```
vgrichina: shouldn't it be chosen from wallet UI?

janedegtiareva: We need to know accountId on the shell side, so it's either passing accountId to wallet and then failing if it can't be used, or setting up some callback mechanism to shell. 1st option seems much simpler

janedegtiareva:
> 1. Are there any corresponding wallet changes? Which branch?
I haven't pushed it yet, but they are going to be basically cosmetic. (supporting not having successUrl + diff permissions)
> 2. Is there a plan to offer user to login when executing other commands without having a valid key?
Out of scope for this change, but it's a good user story to add to the backlog.

bin/near
vgrichina: 1st option isn't good UX for when you don't have any account yet.

It makes sense though as a shorthand to pre-select existing account in wallet.

vgrichina:
> * Are there any corresponding wallet changes? Which branch?
>   I haven't pushed it yet, but they are going to be basically cosmetic. (supporting not having successUrl + diff permissions)

Anyway it makes sense to review them together with this PR.

> * Is there a plan to offer user to login when executing other commands without having a valid key?
>   Out of scope for this change, but it's a good user story to add to the backlog.

AFAIK we discussed this as a basic part of UX for the feature. Isn't a blocker for this PR, but is necessary to provide decent UX.

janedegtiareva:
added the prompt to enter account id. callback to shell would be a diff pr

index.js
janedegtiareva: done

index.js
```diff
@@ -105,3 +106,27 @@ exports.stake = async function(options) {
     const result = await account.stake(options.publicKey, BigInt(options.amount));
     console.log('Result: ', JSON.stringify(result));
 }
+
+exports.login = async function(options) {
+    if (!options.walletUrl) {
+        console.log("Log in is not needed on this environment. Please use appropriate master account for shell operations.")
```
vgrichina: this is hard to understand for outsider. Probably makes sense to link to docs on how to run local setup.

index.js
```diff
@@ -105,3 +106,27 @@ exports.stake = async function(options) {
...
+    } else {
+        const newUrl = new URL(options.walletUrl + "/login/");
+        const title = 'NEAR Shell';
+        newUrl.searchParams.set('title', title);
+        const keyPair = await KeyPair.fromRandom('ed25519');
+        newUrl.searchParams.set('public_key', keyPair.getPublicKey());
+        console.log(`Please navigate to this url and follow the insturctions to log in: ${newUrl.toString()}`);
```
vgrichina: s/insturctions/instructions/

probably good idea to put newline before URL to make it easier to copy if not clickable in given terminal

PR #62, opened 2019-07-09
Add tx-status command to lookup transaction status by hash

vgrichina:


vgrichina:
> Optional: add some text about how to get transaction hash to the command description?

added note about error messages + added hash to error where it was missing:
https://github.com/nearprotocol/nearlib/commit/9939d35e0bfcfc650990f46bbcd00ecb9f61794e

PR #65, opened 2019-07-17
Add require for url so that near login works on older node versions +…

janedegtiareva:
… fix  typo


index.js
```diff
@@ -116,6 +117,7 @@ exports.stake = async function(options) {
 }
 
 exports.login = async function(options) {
+    console.log(options);
```
vgrichina: looks like leftover from debugging

PR #68, opened 2019-07-17
npm version bump

janedegtiareva:


vgrichina:
please next time use `np --yolo` if you need to publish despite failing tests

PR #75, opened 2019-07-23
Hotfix for blank project to work with nightshade api

janedegtiareva:


janedegtiareva:
This will let people work with new_project right away, but a more appropriate fix is to switch to a wallet-integrated example. (This will make local workflow more complex, so I want to discuss this first)

PR #83, opened 2019-07-29
enabling creating test users for apps while running local node 

janedegtiareva:


blank_project/src/index.html
```diff
@@ -1,19 +1,38 @@
...
+          <button id="say-hi" class="btn btn-success btn-lg btn-block ">Say "Hi!"</button>
+        </div>
+        <div class="col-sm-3">
+          <button id="sign-out-button" class="btn btn-danger btn-lg btn-block">Sign-out</button>
+        </div>
+      </div>
+    </div> 
   </div>
-  <script src="https://cdn.jsdelivr.net/npm/nearlib@0.10/dist/nearlib.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/nearprotocol/nearlib@82eed3127d56984e850cf1581344bc098b9f83ab/dist/nearlib.js"></script>
```
vgrichina: should use npm version now that it's live

blank_project/src/config.js
```diff
@@ -18,6 +18,8 @@
                     networkId: 'local',
                     nodeUrl: 'http://localhost:3030',
                     keyPath: '~/.near/validator_key.json',
+                    walletUrl: 'http://localhost:4000',
```
vgrichina: I suggest moving `_wallet` to something like `_wallet/login`, then whole `walletUrlSuffix` thing won't be needed.

blank_project/src/_wallet
```diff
@@ -0,0 +1,31 @@
...
+      <code id="shell-command"></code>
+  </div>
+  <input type="text" id="accountId" name="accountId" placeholder="Account id"></input>
+  <button type="button" onClick="done()">done</button>
+  <script src="https://cdn.jsdelivr.net/gh/nearprotocol/nearlib@82eed3127d56984e850cf1581344bc098b9f83ab/dist/nearlib.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/js-cookie@2/src/js.cookie.min.js"></script>
+  <script src="./config.js"></script>
+  <script>
+    const currentUrl = new URL(window.location.href);
+    const message = `near create_account {newAccountId} --masterAccount {masterAccountId} --publicKey ${currentUrl.searchParams.get('public_key')} --initialAmount 1000`;
```
vgrichina: `--masterAccount {masterAccountId}` – my understanding this shouldn't normally be needed cause obtained from nearcore?

janedegtiareva: It's a good option to spell out in case they want to use a different account to transfer funds from.

blank_project/src/config.js
janedegtiareva: It needs to be /wallet/login/?title=...

vgrichina: > It needs to be /wallet/login/?title=...

so yes, you can put `login` file into `wallet` folder and you'll get this.  Anything I'm missing?

blank_project/src/_wallet
vgrichina: then it makes sense to have 2 different suggestions – one with all possible params, and one minimal one which would easy to copy-paste and run (editing command line in multiple places ins't fun).

my understanding is that this should be enough for most cases:
```
`near create_account --publicKey ${currentUrl.searchParams.get('public_key')} {newAccountId}`
```

blank_project/src/config.js
janedegtiareva: If you do that and run npm run start, http://localhost:4000/wallet/login/?(needed url params)  doesn't work

blank_project/src/_wallet
janedegtiareva: @potatodepaulo  dev UX question: 2 commands to choose from with an easy copy-paste option vs 1 command that needs to be tweaked -- what do you think is better?

blank_project/src/config.js
janedegtiareva: updated to use /wallet/login/index.html which works with the current url template

blank_project/src/_wallet
janedegtiareva: For context, it's a diff between these 2 (roughly)
Screen Shot 2019-07-30 at 4 16 06 PM
Screen Shot 2019-07-30 at 4 19 13 PM

potatodepaulo: Adding two cents here: 
The one line is much less confusing to an outsider coming in. We should always reduce the number of choices a user must make. I think the long term solution is that the command is fully generated, but this is a step in that direction.

potatodepaulo: @vgrichina in this specific case, I think here is the solution: 
We show the long command so that the user doesn't need to think about what they're copying and pasting. Even in the cases this is more than needed, that use case is still not going to break.
To address your point we should add a link here on this page that says "What am I seeing?" that takes them into the shell documentation (Or the new accounts documentation we're working on). This is a good compromise, since savvy users will be able to figure out their own commands.

janedegtiareva:
After some discord conversations looks like we can move forward with the latest version. I'll make a NEP for wallet url scheme.

potatodepaulo:
> Approved conditional on merge conflict fixes and rebasing.

Adding for future notes 💯

PR #85, opened 2019-07-31
Upgrade blank project to gulp 4 so that it works with latest node

janedegtiareva:
Tested with blank_project on testnet (deploys app successfully)

### DIFF
--- a/blank_project/gulpfile.js
+++ b/blank_project/gulpfile.js
@@ -1,30 +1,20 @@
 const gulp = require("gulp");
 const nearUtils = require("near-shell/gulp-utils");
 
-gulp.task("build:model", callback => {
-  nearUtils.generateBindings("model.ts", "../out/model.near.ts", callback);
-});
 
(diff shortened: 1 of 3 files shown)


vgrichina:
@janedegtiareva do you plan to update studio or should I do that?

blank_project/gulpfile.js
```diff
@@ -1,30 +1,20 @@
...
 const nearUtils = require("near-shell/gulp-utils");
 
-gulp.task("build:model", callback => {
-  nearUtils.generateBindings("model.ts", "../out/model.near.ts", callback);
-});
 
-gulp.task("build:bindings", ["build:model"], callback => {
-  nearUtils.generateBindings("main.ts", "../out/main.near.ts", callback);
-});
+function build_bindings(done){
```
ilblackdragon: just a small thing, but let's keep function names consistent - generateBindingsMain and generateBindingsModel?

PR #86, opened 2019-07-31
Return default blank project config if there is no config in expected…

janedegtiareva:
… dir


janedegtiareva:
Fix for https://github.com/nearprotocol/near-shell/issues/77

PR #89, opened 2019-08-01
Switch to package management

willemneal:
Currently all near related deps are a branch on my repos.  So once those get pulled we can merge this one.


vgrichina:
@willemneal make sure we also have corresponding studio changes before merging

willemneal:
Yeah I just thought I'd start these PRs so that we'd have the code to test it with.  For each one I have a corresponding `#near_dist` branch so that you can add them as dependencies when testing studio.

janedegtiareva:
Do you mind adding some context to the PR description with some notes on what will change with this PR?

blank_project/assembly/main.ts
```diff
@@ -1,6 +1,4 @@
-import { context, storage, near } from "./near";
-
-import { Greeter } from "./model.near";
+import { Greeter } from "../out/model.near";
```
vgrichina: @willemneal will this work fine when `./model` has some dependencies within project?

gulp-utils.js
```diff
@@ -3,7 +3,7 @@ function generateBindings(inputFile, outputFile, callback) {
   const asc = getAsc();
   asc.main([
     inputFile,
-    "--baseDir", "assembly",
+    "--baseDir", process.cwd(),
```
vgrichina: it's likely not a thing in browser (when running in studio)

gulp-utils.js
```diff
@@ -40,36 +39,13 @@ function getAsc() {
     }
 
     const logLn = process.browser ? window.logLn : console.log;
+    console.log("asc " + args.join(" "));
```
vgrichina: debug log leftover

package.json
```diff
@@ -23,7 +23,7 @@
   },
   "devDependencies": {},
   "dependencies": {
-    "assemblyscript": "github:nearprotocol/assemblyscript#assemblyscript-update",
+    "assemblyscript": "github:willemneal/assemblyscript#near_dist",
```
vgrichina: Please make branch in nearprotocol repo, I've added you as collaborator.

bowenwang1996: We should setup CI for external contributors as well. Gitlab CI does not work for that

vgrichina: @bowenwang1996 what does work? Travis?

bowenwang1996: Assemblyscript uses travis and it works

gulp-utils.js
willemneal: You could just mock it, at least that's what [BrowserFS](https://github.com/jvilk/browserfs) does.

blank_project/assembly/main.ts
willemneal: With the current setup you have to create the near file from model, so I'm not exactly sure.  I think to make that work, we either have to output the near files into the same directory as the source or copy over all of the imports.

I really want to remove the step of having to create temporary files from the tool chain.  I posed an idea to Bowen of using decorators on classes to tell the compiler that the class should be transformed during compilation.

willemneal: Looks like you already use `process.cwd()` in the same file:
```
readFile: (filename, baseDir) => {
        baseDir = pathModule.relative(process.cwd(), baseDir);
```

blank_project/package.json
```diff
@@ -12,14 +12,14 @@
     "test": "npm run build && jest test --env=near-shell/test_environment"
   },
   "devDependencies": {
-    "assemblyscript-json": "github:nearprotocol/assemblyscript-json",
+    "assemblyscript-json": "github:WebAssemblyOS/assemblyscript-json#near_dist",
```
janedegtiareva: Is the plan going forward to switch to this branch instead of nearprotocol fork?

willemneal: No this was just because I was waiting on the PR since it fixes its exports.

PR #92, opened 2019-08-01
Fix sending money (added an explit param for account to send from)

janedegtiareva:


janedegtiareva:
Tested on local shell. I want to add a test for this, but there are higher priority things on my plate right now.

index.js
```diff
@@ -97,9 +97,10 @@ exports.scheduleFunctionCall = async function(options) {
 };
 
 exports.sendTokens = async function(options) {
```
potatodepaulo: Nit: Should we change this to match API?

potatodepaulo: e.g. `exports.sendMoney`

PR #95, opened 2019-08-02
Update main.ts in blank template (this did not make it in for origina…

janedegtiareva:
…l update)


janedegtiareva:
contract builds and deploys to local node. view call works (near view testcontract3 whoSaidHi)

PR #98, opened 2019-08-02
Improve error handling for login

janedegtiareva:


janedegtiareva:
Previously, if there was a server error while querying account, the process failed with a scary stack trace as output. Now it logs a retry message.

PR #101, opened 2019-08-03
Provide human readable response for schedule function calls

janedegtiareva:


janedegtiareva:
Fix for https://github.com/nearprotocol/near-shell/issues/82
Now outputs, e.g. 
[janecontract2]: LOG: janetestnet says "Hi!"
Result: Hello

PR #115, opened 2019-08-16
Add null check to string returned by export

willemneal:


vgrichina:
Looks like this isn't needed anymore

PR #128, opened 2019-09-01
Add interactive `near repl` with auto-injected `nearlib`, `near` and `account`

vgrichina:
Needed this for my own debugging, I think will be useful for everyone else as well


bin/near
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --experimental-repl-await
```
janedegtiareva: what's --experimental-repl-await?

janedegtiareva: (would be helpful to leave a comment)

janedegtiareva:
looks like it needs some changes for ci (it fails with /usr/bin/env: 'node --experimental-repl-await': No such file or directory)

bin/near
vgrichina: seems to be very googlable, so probably not worth it:

https://www.google.com/search?client=safari&rls=en&q=--experimental-repl-await&ie=UTF-8&oe=UTF-8

vgrichina:
> looks like it needs some changes for ci (it fails with /usr/bin/env: 'node --experimental-repl-await': No such file or directory)

yeah, it's kinda weird, cannot reproduce this locally. Both BSD `env` and coreutils `env` are supposed to support this syntax.

Maybe it should be fixed together with https://github.com/nearprotocol/near-shell/issues/45 (as then there is reason for us to launch child process).

PR #133, opened 2019-09-05
React template add

icerove:
create new react template for developers. Command is near new_react_project


icerove:
Screen Shot 2019-09-05 at 5 49 30 PM
this is the page after steps in readme. and a very simple function in the contract

vgrichina:
@icerove  @potatodepaulo @ilblackdragon @janedegtiareva

### meta-comment on CLI UX:

I think given we have more than one template now, it's good time to switch to pattern used by other projects (e.g. create-react-app which is one of most popular ways to create a React app):

https://github.com/facebook/create-react-app#yarn

Basically then if I have `yarn` (and don't have any NEAR tools) I can just do:

```
yarn create near-react-app my-awesome-app
```

or if I don't even have `yarn`:

```
npx create-near-react-app my-awesome-app
```

### on UI:

Jake have designed brand site, we should use fonts/colors from there I think:
https://nearprotocol.com/brand/

blank_react_project/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+Step 3:
+deploy the smart contract, run the command:
+```
+npm run deploy:contract
+```
+Step 4:
+Finally, run the command in your terminal.
+```
+npm(yarn) intall && npm(yarn) start
```
vgrichina: it's better to have copy-pastable commands, i.e. in this case if you want to provide alternate just spell them out separately like this:

```
yarn && yarn start
```

```
npm install && npm run start
```

vgrichina: though I'd prefer we just stick to `yarn` everywhere to avoid confusion, unless someone has objections

blank_react_project/assembly/main.ts
```diff
@@ -0,0 +1,15 @@
+import { near, context, storage, logging, base58, base64, PersistentMap, PersistentVector, PersistentDeque, PersistentTopN, ContractPromise, math } from "near-runtime-ts";
+// import { u128 } from "bignum";
+// import { TextMessage } from "./model";
+
+const NAME = ". Welcome to NEAR Protocol chain"
+
+export function hello(name: string): string {
```
vgrichina: looks like we have both 2-space and 4-space indentation in different places of PR. I think we should stick to same indent size everywhere for simplicity, 4-space seems to match most of our stuff (unless anyone has objections)

blank_react_project/assembly/model.ts
```diff
@@ -0,0 +1,4 @@
+export class TextMessage {
+    sender: string;
+    text: string;
+  }
```
vgrichina: formatting issue

blank_react_project/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+        viewMethods: [
+            "hello",],
+        changeMethods: [
+            ],
+        sender: window.accountId
+    });
+}
+
+window.nearInitPromise = doInitContract().then(() => {
+    ReactDOM.render(<App contract={window.contract} wallet={window.walletAccount}/>,
```
vgrichina: app rendering shouldn't be delayed to until contract is setup as then user won't see app UI until it happens, which is undesirable

index.js
```diff
@@ -31,6 +31,19 @@ exports.newProject = async function(options) {
   console.log('Copying project files complete.')
 };
 
+exports.reactProject = async function(options) {
```
vgrichina: it's exact copy of `newProject`, better to pass what's different (`sourceDir`) as a parameter than copy whole method.

janedegtiareva:
> @icerove @potatodepaulo @ilblackdragon @janedegtiareva
> 
> ### meta-comment on CLI UX:
> I think given we have more than one template now, it's good time to switch to pattern used by other projects (e.g. create-react-app which is one of most popular ways to create a React app):
> 
> https://github.com/facebook/create-react-app#yarn
> 
> Basically then if I have `yarn` (and don't have any NEAR tools) I can just do:
> 
> ```
> yarn create near-react-app my-awesome-app
> ```
> 
> or if I don't even have `yarn`:
> 
> ```
> npx create-near-react-app my-awesome-app
> ```
> 
> ### on UI:
> Jake have designed brand site, we should use fonts/colors from there I think:
> https://nearprotocol.com/brand/

Should we support both methods for creating a new project (yarn create and near new_project)? Some folks who are less familiar with javascript environment might prefer that @nearmax

vgrichina:
@janedegtiareva 
> Some folks who are less familiar with javascript environment might prefer that @nearmax

someone who's not familiar with JS env will be copy-pasting whatever we specify in docs (and seems like it's less confusing if we document only one way of doing it).

nearmax:
> someone who's not familiar with JS env will be copy-pasting whatever we specify in docs (and seems like it's less confusing if we document only one way of doing it).

I doubt that people who will be writing financial contracts on our system (like some derivates startups) will be mindlessly copy-pasting yarn commands beyond their first tutorial. As soon as they get really large money involved they would want understanding of what they do with our blokchain, like what are the commands they issuing.

On the related topic that we have already discussed it with @janedegtiareva : currently Rust smart contract developers have to work in JS-oriented environment because our UX is JS-centric. Specifically, operations with near-shell require creating a directory that contains various JS-specific files even if people work with Rust only. This seems to be a subpar UX. Remember a similar situation that we had in the past when in NearStudio we were showing to the users all kind of std files and helpers and that was confusing to them and so we hid it to optimize for UX? We have a similar situation now -- JS-files are not related to what Rust contract developers are doing and therefore they should not be created if Rust user works with near-shell. WDYT? CC @ilblackdragon

vgrichina:
> I doubt that people who will be writing financial contracts on our system (like some derivates startups) will be mindlessly copy-pasting yarn commands beyond their first tutorial.

I have no doubt most devs will be copy-pasting stuff without much thinking. Maybe someone will do security audit later if it involves huge money. But people will definitely just copy-paste stuff when they are trying out our stuff (and that's most important stage for UX as then they aren't sold on our stuff yet).

> On the related topic that we have already discussed it with @janedegtiareva : currently Rust smart contract developers have to work in JS-oriented environment because our UX is JS-centric. 

That's an argument for following what I suggest: separate JS-specific templates into their own repos and have `near-shell` only deal with providing CLI for node.

Rust has it's own tools for project templating as well (e.g. https://github.com/ashleygwilliams/cargo-generate) and it makes sense to use them for Rust projects.

nearmax:
I am all for separating templating functionality from near-shell.

potatodepaulo:
> I am all for separating templating functionality from near-shell.

I second this. Also, I think that creating a template for people to create templates within these other tools is the easiest thing to maintain. I imagine people might want to scaffold a react project as much as a Vue project. I'm sure there are different types of applications in Rust that people would need different libraries for as well.

nearmax:
Created tracking issue: https://github.com/nearprotocol/near-shell/issues/144

blank_react_project/README.md
icerove: sorry not get any notification until today I check it myself. I will change it

icerove: for using fonts/colors from brand, where I can find the asserts?

icerove: change read me with separately copyable command

blank_react_project/assembly/model.ts
icerove: change it

blank_react_project/src/index.js
icerove: but it has to be contract ready and then to render dom, so I think it is okey since corgi also present like this and no blank UI happen

index.js
icerove: since we are going to seperate repo, think of not doing that

icerove:
this is how it like for now: 
Screen Shot 2019-09-24 at 11 43 18 AM

vgrichina:
@icerove is this fully migrated to https://github.com/nearprotocol/create-near-app? Can we close this PR?

icerove:
Yes, close it

PR #137, opened 2019-09-09
Remove confusing "failed to load validator key" warning.

janedegtiareva:
This happens because we try to load the validator key for all environments


janedegtiareva:
this is a fix for https://github.com/nearprotocol/near-shell/issues/134

PR #139, opened 2019-09-10
Use inspectResponse to display nearcore responses

vgrichina:


janedegtiareva:
Could you check the test failure?

PR #140, opened 2019-09-11
Fix mkdirp

willemneal:
- Add fs methods check so that it has default behavior in NEARStudio, since the write file function handles mkdirp

### DIFF
--- a/blank_project/package.json
+++ b/blank_project/package.json
@@ -12,8 +12,6 @@
     "test": "npm run build && jest test --env=near-shell/test_environment"
   },
   "devDependencies": {
-    "assemblyscript-json": "github:nearprotocol/assemblyscript-json",
-    "bignum": "github:MaxGraey/bignum.wasm",
     "gh-pages": "^2.1.1",
     "gulp": "^4.0.2",
     "jest": "^24.8.0",
--- a/gulp-utils.js
+++ b/gulp-utils.js
@@ -60,10 +60,15 @@ function getAsc() {
       },
       writeFile: (filename, contents) => {
         const name = filename.startsWith("../") ? filename.substring(3) : filename;
-        mkdirp(name);
+        if (fs.mkdirSync){
+          mkdirp(name);
+        }
         fs.writeFileSync(name, contents);
       },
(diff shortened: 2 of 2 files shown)


blank_project/package.json
```diff
@@ -12,8 +12,6 @@
     "test": "npm run build && jest test --env=near-shell/test_environment"
   },
   "devDependencies": {
-    "assemblyscript-json": "github:nearprotocol/assemblyscript-json",
```
vgrichina: why are these getting removed?

gulp-utils.js
```diff
@@ -50,17 +56,30 @@ function getAsc() {
...
         const name = filename.startsWith("../") ? filename.substring(3) : filename;
+        if (fs.mkdirSync){
+          mkdirp(name);
+        }
         fs.writeFileSync(name, contents);
       },
-      listFiles: () => []
+      listFiles: (dirname, baseDir) => {
+        if (fs.readdirSync == undefined){
+          return [];
```
vgrichina: hmm, if it's ok to remove empty list from `listFiles` and it doesn't break anything – why do we need real implementation at all?

gulp-utils.js
```diff
@@ -31,9 +21,25 @@ function getAsc() {
   }
 
   asc = require("assemblyscript/bin/asc");
-
+  
   const fs = require("fs");
   const pathModule = require("path");
+
+  // Create parent directories if they don't exist
+  function mkdirp(path){
```
vgrichina: it doesn't really work like `mkdir -p` (e.g. if you pass `foo/bar.txt` you'd expect `bar.txt` directory created by `mkdir-p`), so either logic or name should be changed

gulp-utils.js
```diff
@@ -1,13 +1,3 @@
-// FUTURE PEOPLE: This file is called "gulp-utils" but it's not related to the deprecated library called "gulp-utils". Don't panic.
```
vgrichina: I think `// FUTURE PEOPLE` part should stay

janedegtiareva:
@willemneal what would you like to do with this PR?

PR #143, opened 2019-09-17
Add delete account + test to near shell

janedegtiareva:


test/index.sh
```diff
@@ -2,4 +2,4 @@
 set -ex
 ./test/new_project.sh
 ./test/account_operations.sh
-./test/contract_tests.sh
+#./test/contract_tests.sh
```
vgrichina: is this intentional?

vgrichina: nevermind, saw in history

PR #146, opened 2019-09-18
Remove unneeded deps and pin near-runtime for testing

willemneal:
Also fixed test script to install local version of near-shell

### DIFF
--- a/blank_project/package.json
+++ b/blank_project/package.json
@@ -12,8 +12,6 @@
     "test": "npm run build && jest test --env=near-shell/test_environment"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


willemneal:
@vgrichina  It builds now, but fails other tests with nearcore.  There are still failures with nearcore in the runtime PR too.

janedegtiareva:
> @vgrichina It builds now, but fails other tests with nearcore. There are still failures with nearcore in the runtime PR too.

What's the error message?

vgrichina:
merged this with some updates https://github.com/nearprotocol/near-shell/commit/46e57a17240d12084555f07cab34c29e16bb6079

this PR for some reason was on `inspect-response` branch, so I rebased

PR #149, opened 2019-09-23
Change input to near tokens instead of attonear

janedegtiareva:


vgrichina:
meta-comment: I'm pretty sure we still need to support amounts smaller than 1 NEAR

janedegtiareva:
> meta-comment: I'm pretty sure we still need to support amounts smaller than 1 NEAR

Yeah, this PR is not ready for the review yet. I am testing various use cases now and running into doc troubleshooting slow downs.

PR #150, opened 2019-09-24
Use new near-runtime in blank_project

willemneal:
- Update blank project
- Keep gulp-utils with the same interface


vgrichina:
anything touching assemblyscript/gulp-utils generally needs to be tested in studio as well (so corresponding studio PR is expected)

It's also IMO will be less confusing to avoid reusing `staging` branch, but create branch corresponding to single topic and use it in all projects containing related changes. So that it's clear what branches need to be merged for a given change to go live.

willemneal:
Yeah having the staging branch makes it easier to make the other repos depend on the PR.   We could also move to adding https://www.pika.dev/cdn to each repo, which then let's you depend on a specific PR and updates after each commit.

I also was going to make a separate PR to Studio as well just waiting for the green check mark here.  I also am looking into the floating point issue.

vgrichina:
how exactly does `staging` branch make it easier to depend on the PR? Like isn't it literally as easy to depend on any other branch?

PR #151, opened 2019-09-24
Add eslint and fix style issues

vgrichina:
This is to make code style consistent with nearlib

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+tmp-project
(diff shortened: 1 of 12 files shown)


vgrichina:
> I thought in Javascript/Typescript people prefer 2 space indentation.

a lot of peeps prefer that way, but that's been mostly driven by callback hell era (before `async` / `await`) when stuff had to be super-nested.

PR #152, opened 2019-09-24
Use near-assemblyscript 0.7.3

vgrichina:


willemneal:
The issue is that the gulp file still depends on `assemblyscript` not `near-assemblyscript`:
https://github.com/nearprotocol/near-shell/blob/69af793108905465ecd443bbb7d6e4349c5fbfd1/gulp-utils.js#L33

PR #154, opened 2019-09-26
Use new near-runtime in blank_project 

willemneal:
Replaces #150


willemneal:
@vgrichina Not sure why this is failing.

PR #160, opened 2019-10-11
Set initialBalance for create_account - 1 NEAR

vgrichina:


vgrichina:
Merging as passes on staging (master CI needs to update test.near balance)

PR #178, opened 2019-11-04
Add `near generate-key` command

vgrichina:
Fixes https://github.com/nearprotocol/near-shell/issues/177

### DIFF
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -11,7 +11,7 @@ module.exports = {
             const { deps: { keyStore }} = near.config;
             const existingKey = await keyStore.getKey(argv.networkId, argv.accountId);
             if (existingKey) {
-                console.log(`Account has existing key pair with ${existingKey.publicKey} public key`)
+                console.log(`Account has existing key pair with ${existingKey.publicKey} public key`);
             } else {
                 const keyPair = KeyPair.fromRandom('ed25519');
                 await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -9,9 +9,14 @@ module.exports = {
         let near = await require('../utils/connect')(argv);
         if (argv.accountId) {
             const { deps: { keyStore }} = near.config;
-            const keyPair = KeyPair.fromRandom('ed25519');
-            await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
-            console.log(`Generated key pair with ${keyPair.publicKey} public key`);
+            const existingKey = await keyStore.getKey(argv.networkId, argv.accountId);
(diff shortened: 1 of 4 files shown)


commands/generate-key.js
```diff
@@ -0,0 +1,17 @@
...
+    handler: exitOnError(async (argv) => {
+        let near = await require('../utils/connect')(argv);
+        if (argv.accountId) {
+            const { deps: { keyStore }} = near.config;
+            const keyPair = KeyPair.fromRandom('ed25519');
+            await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
+            console.log(`Generated key pair with ${keyPair.publicKey} public key`);
+        }
+    })
+};
```
icerove: why do we have this function individually not with other functions? and this makes command and function together? why not follow what we already have?

vgrichina: @icerove didn't want to do too many chnages at once to pull every function, so for now only pulled new functions like this

The reason to do this is that it's more convenient to work with such structure if there are concurrent modifications that needs to be merged (e.g. we both add some commands).
Also current structure avoids extra step (you just define a command at once, without having to define a function first and then wrap it into command).

vgrichina:
> Looks like this will override the old key. Is that the best option?

@bowenwang1996 do you think we should ask the user whether to overwrite old key? Or just no-op with warning if key does exists?

bowenwang1996:
I think we should ask user before overriding the key file because once done, the private key would be gone. Normally that is fine because it's the same account, but if they actually want to use that key for a different network, they will lose their old key and that actually matters.

vgrichina:
@bowenwang1996 the problem you described is solved by doing no-op with warning too though. Benefit over asking is that it's easier to put in a script (does not require user input).

BTW we also can add separate network ID for stake wars.

bowenwang1996:
Sorry I don't understand. If you override your old key and that is your only key to your account, it seems that you will lose your account. Did I miss something?

icerove:
How about only generate new public keys to that account and do not change account private key? not generate key pair but only public one. Since multi public keys are accepted.

vgrichina:
> How about only generate new public keys to that account and do not change account private key? not generate key pair but only public one. Since multi public keys are accepted.

@icerove hmm, do you mean deriving multiple different key pairs from one private key? Sounds like a bit out of scope.

@bowenwang1996 
> Sorry I don't understand. If you override your old key and that is your only key to your account, it seems that you will lose your account. Did I miss something?

Yes, that I am suggesting to solve this problem in a different way: make near-shell not do anything when there is existing key for account, but just show a warning message. This also requires changes to PR, just a different change.

bowenwang1996:
Oh I see. Sure that works.

PR #189, opened 2019-11-15
Fix tx-status to handle account id better

vgrichina:
Fixes https://github.com/nearprotocol/near-shell/issues/187

### DIFF
--- a/commands/tx-status.js
+++ b/commands/tx-status.js
@@ -13,7 +13,7 @@ module.exports = {
             required: true
         }),
     handler: exitOnError(async (argv) => {
-        const near = await connect(argv)
+        const near = await connect(argv);
 
         const hashParts = argv.hash.split(':');
         let hash, accountId;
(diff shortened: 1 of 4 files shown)


icerove:
It is really a goo idea to separate the command when you fix it.
Please also update the readme, I think it is not including the new command you added.

commands/tx-status.js
```diff
@@ -0,0 +1,37 @@
...
+        if (hashParts.length == 2) {
+            [accountId, hash] = hashParts;
+        } else if (hashParts.length == 1) {
+            [hash] = hashParts;
+        } else {
+            throw new Error('Unexpected transaction hash format');
+        }
+        accountId = accountId || argv.accountId || argv.masterAccount;
+
+        if (!accountId) {
```
icerove: in this command, accountId is actually not required and user will only follow the command we give not given the account id by default? Maybe change instructions in the readme for this since if not accountId, there will throw an error.

vgrichina:
> It is really a goo idea to separate the command when you fix it.

Seemed like good opportunity to do this. If command was more complex I'd split the change in 2 separate (first move, then fix).

> Please also update the readme, I think it is not including the new command you added.

I didn't add any new command, not sure what do you mean.

icerove:
near generate-key and near repl is not added. Also, still not clear about accountid problem I left above?

vgrichina:
> near generate-key and near repl is not added.

these aren't part of this PR

commands/tx-status.js
vgrichina: the intent is to change other tools to return transaction hash as `accountId:hash`, so we can just have one string that can be copy-pasted

see https://github.com/nearprotocol/near-shell/issues/187

vgrichina:
> near generate-key and near repl is not added. Also, still not clear about accountid problem I left above?

https://github.com/nearprotocol/near-shell/commit/1c139dd33cb5ec0cd15cbf8e6070efdc21e8d753

PR #192, opened 2019-11-18
Update to use asc v0.8.0 and near-bindgen-as

willemneal:
Use near-bindgen-as's compiler api for projects that already use the gulp-utils.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,10 +30,11 @@
     "strip-ansi-cli": "^2.0.0"
   },
   "dependencies": {
+    "assemblyscript": "^0.8.0",
     "bs58": "^4.0.1",
     "jest-environment-node": "^24.5.0",
     "ncp": "^2.0.0",
-    "near-assemblyscript": "^0.7.4",
+    "near-bindgen-as": "github:nearprotocol/near-bindgen-as#v1.1.0",
     "nearlib": "^0.17.1",
     "rimraf": "^3.0.0",
     "yargs": "^14.2.0"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,10 +30,11 @@
     "strip-ansi-cli": "^2.0.0"
   },
   "dependencies": {
+    "assemblyscript": "^0.8.0",
     "bs58": "^4.0.1",
     "jest-environment-node": "^24.5.0",
     "ncp": "^2.0.0",
-    "near-assemblyscript": "^0.7.4",
+    "near-bindgen-as": "github:nearprotocol/near-bindgen-as#v1.1.0",
```
vgrichina: Let's not depend on GitHub version anymore. Just publish it on NPM. Should probably be `0.x.z` version still before main net though.

vgrichina:
1) Needs to not fail CI.
2) Does studio need any corresponding changes?

package.json
willemneal: Yeah I was going to reach out about setting up a github action to publish near-bindgen-as.

vgrichina: I actually prefer explicit manual publish using https://github.com/sindresorhus/np

bowenwang1996:
I think we can merge this when the studio update is also ready @vgrichina

vgrichina:
@bowenwang1996 looks like it's ready https://github.com/nearprotocol/NEARStudio/pull/229

just need to test to make sure all stuff works fine

vgrichina:
@willemneal @bowenwang1996 merged, but we also need to update https://github.com/nearprotocol/create-near-app templates now I think.

willemneal:
I think you just need to update it to use the new near-shell since the
gulp-utils just uses the bindgens compiler interface.

PR #193, opened 2019-11-18
add version check (after any command) fixes #183

### LINKED ISSUE #183: runtime version check
can we add "most recent version check"? 

today i ran `near-shell` and then stopped to wonder which version i was on -- it was a few weeks old. this is especially important while we're making high frequency changes to this repo.

the current version (while writing this issue) is `0.14.1`

```
➜  dev npm list -g | grep near
├─┬ near-shell@0.11.3
│ ├─┬ near-assemblyscript@0.7.4
│ ├─┬ nearlib@0.14.1
```

(as an aside, it confused me for a second to notice that `nearlib` just happens to be at _exactly_ the same version number, although the two are not synchronized in their versioning.)

```
➜  dev npm outdated -g
Package     Current  Wanted  Latest  Location
near-shell   0.11.3  0.11.5  0.14.1  global
```

amgando:
added update-notifier with cleaner messages for the user.  this requires to the user actually attempt to issue a command.  given current config of yargs it *does not always* print out the need for an upgrade, only when a command is added.

note also, **i didn't add any tests for this**

maybe these links will help with follow up

testing `yargs` output
https://github.com/yargs/yargs/blob/master/docs/advanced.md#testing-a-command-module

(offtopic) also consider using nave for testing against node versions
https://github.com/isaacs/nave

### DIFF
--- a/utils/check-version.js
+++ b/utils/check-version.js
@@ -8,13 +8,7 @@ const chalk = require('chalk');  // colorize output
 const isCI = require('is-ci');   // avoid output if running in CI server
 
 // updateCheckInterval is measured in seconds
-const ONCE_PER = {
-    SECOND: 0,
-    MINUTE: 1000 * 60,
-    HOUR:   1000 * 60 * 60,
-    DAY:    1000 * 60 * 60 * 24,
-    WEEK:   1000 * 60 * 60 * 24 * 7
-};
+const UPDATE_CHECK_INTERVAL_SECONDS = 1;
 
 /**
     check the current version of NEAR Shell against latest as published on npm
@@ -22,7 +16,7 @@ const ONCE_PER = {
 module.exports = async function checkVersion() {
     const pkg = require('../package.json');
 
-    const notifier = updateNotifier({ pkg, updateCheckInterval: ONCE_PER.SECOND });
+    const notifier = updateNotifier({ pkg, updateCheckInterval: UPDATE_CHECK_INTERVAL_SECONDS });
 
     if (
         notifier.update &&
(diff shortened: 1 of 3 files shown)


utils/check-version.js
```diff
@@ -0,0 +1,75 @@
...
+    https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-cli
+    https://github.com/npm/cli
+ */
+const updateNotifier = require('update-notifier');
+const chalk = require('chalk');  // colorize output
+const isCI = require('is-ci');   // avoid output if running in CI server
+
+// updateCheckInterval is measured in seconds
+const ONCE_PER = {
+    SECOND: 0,
```
vgrichina: shouldn't be 1 instead of 0?

utils/check-version.js
```diff
@@ -0,0 +1,75 @@
...
+    MINUTE: 1000 * 60,
+    HOUR:   1000 * 60 * 60,
+    DAY:    1000 * 60 * 60 * 24,
+    WEEK:   1000 * 60 * 60 * 24 * 7
+};
```
vgrichina: let's just keep single constant, e.g. `UPDATE_CHECK_INTERVAL_SECONDS`

amgando:
this fell behind master with conflicts in `package.json` bc of the recent `yargs` version bump.  

should i rebase and force push this branch or is there some other workflow we prefer? 

a guy in the internet says force pushing personal development branches is fine 😄
https://stackoverflow.com/a/15144275/2836874

vgrichina:
> this fell behind master with conflicts in `package.json` bc of the recent `yargs` version bump.
> 
> should i rebase and force push this branch or is there some other workflow we prefer?
> 
> a guy in the internet says force pushing personal development branches is fine 😄
> https://stackoverflow.com/a/15144275/2836874

rebase is fine if branch only has your commits, yes

amgando:
forgot to lint 🤕

PR #194, opened 2019-11-18
minor UX fixes

amgando:
remove the extra view command and change wrapping on the text so it adjusts to the terminal size

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -211,7 +211,6 @@ yargs // eslint-disable-line
     .command(deploy)
     .command(scheduleFunctionCall)
     .command(callViewFunction)
-    // .command(viewAccount)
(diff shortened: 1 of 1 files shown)


amgando:
@vgrichina, i have a few more of these waiting but i don't want to add noise to your schedule if this isn't the right time to do this work.  it just happened to be close to the code i was reading so it was easy to edit

bin/near-cli.js
```diff
@@ -211,7 +211,7 @@ yargs // eslint-disable-line
     .command(deploy)
     .command(scheduleFunctionCall)
     .command(callViewFunction)
-    .command(viewAccount)
+    // .command(viewAccount)
```
vgrichina: just remove line completely, no need to keep commented out

PR #199, opened 2019-11-20
add and change command and instruction

icerove:
add near generate-key <account-id>  and near repl .
change near create_account to more clear

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,14 @@ near <command>
 #### For account:
 ```bash
   near login                                       # logging in through NEAR protocol wallet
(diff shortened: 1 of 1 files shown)


vgrichina:
```

hmm, I think `repl` and `generate-key` are already there (I've added after you found the issue)

#### Misc:
  near repl                                    # launch interactive Node.js shell with NEAR connection available to use
  near generate-key <account-id>               # generate key
```

vgrichina:
Closing this as no activity since Nov 20.

PR #201, opened 2019-11-22
Add pretty-printed amount of attonear to near state

janedegtiareva:


janedegtiareva:
Example output {
  amount: '8999999999837087887',
  code_hash: '11111111111111111111111111111111',
  locked: '0',
  storage_paid_at: 852943,
  storage_usage: 182,
  formattedAmount: '~9E attonear'
}

utils/formatting-utils.js
```diff
@@ -0,0 +1,44 @@
...
+*/
+const symbols = ['', 'K', 'M', 'G', 'T', 'P', 'E'];
+
+const symbolsToExponent = {
+    K: 1000,
+    M: 1000 * 1000,
+    G: 1000 * 1000 * 1000,
+    T: 1000 * 1000 * 1000 * 1000,
+    P: 1000 * 1000 * 1000 * 1000 * 1000,
+    E: 1000 * 1000 * 1000 * 1000 * 1000 * 1000,
```
vgrichina: JS can only handle 53-bits of precision for integers https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Global_Objects/Число/MAX_SAFE_INTEGER

So some big int / big decimal library should be used for atto near math (bn.js / big.js / etc).

janedegtiareva: My understanding is that the unsafe behavior is all about rounding errors, which is not an issue if we try to do 6.7E calculation for example.

__tests__/unittest.js
```diff
@@ -0,0 +1,27 @@
...
+    expect(format.prettyPrintNearAmount('99999')).toEqual('99999 attonear'); // do not format smaller values
+    expect(format.prettyPrintNearAmount('8999999999837087887000')).toEqual('~9000E attonear'); 
+});
+
+test('parseInputAmount', async() => {
+    expect(format.parseInputAmount('5K ')).toEqual(5000);
+    expect(format.parseInputAmount('6 M ')).toEqual(6000000);
+    expect(format.parseInputAmount('1.5 K ')).toEqual(1500);
+    expect(format.parseInputAmount('9000E')).toEqual(9000000000000000000000); 
+    expect(format.parseInputAmount('8999999999837087887000')).toEqual('8999999999837087887000');
```
ilblackdragon: we for sure need here test with 10^27 (e.g. 1B nears)

utils/formatting-utils.js
```diff
@@ -0,0 +1,44 @@
+/*
+1 K = 1000 B, K for 3 zero
```
ilblackdragon: 1000 B?

utils/formatting-utils.js
```diff
@@ -0,0 +1,44 @@
+/*
```
ilblackdragon: generally, should the formatting be the actual prefixes:
<10^3 => attoNEAR
<10^6 => femtoNEAR
<10^9 => picoNEAR
> 10^12 => nanoNEAR
< 10^15 => 1 microNEAR
< 10^18 => milliNEAR
rest => NEAR

utils/formatting-utils.js
ilblackdragon: Generally we should never operate in floats. 

There are not guarantees around precision around floats across platforms.
For example on my laptop:
```
> 6.7 == 6.7000000000000000001
true
```

We are already using bn.js everywhere, so just using it here makes a lot of sense and doesn't open possible issues with rounding.

__tests__/unittest.js
ilblackdragon: and also check non Equal as well, because 
```
> 1000000000000000001 === 1000000000000000000
true```

utils/formatting-utils.js
janedegtiareva: Good question. This version came from validator feedback, but either would work. https://github.com/nearprotocol/stakewars/issues/36

janedegtiareva:
> Example output {
> amount: '8999999999837087887',
> code_hash: '11111111111111111111111111111111',
> locked: '0',
> storage_paid_at: 852943,
> storage_usage: 182,
> formattedAmount: '~9E attonear'
> }
The new format, based on a survey, is to use NEAR as the standard way to input amounts (consistent with the wallet)

__tests__/unittest.js
janedegtiareva: done

utils/formatting-utils.js
janedegtiareva: added BN

frol:
@janedegtiareva @vgrichina How do you handle the transition from attoNEAR to yoctoNEAR given that https://github.com/nearprotocol/nearcore/pull/1827 is not merged yet?

vgrichina:
> @janedegtiareva @vgrichina How do you handle the transition from attoNEAR to yoctoNEAR given that [nearprotocol/nearcore#1827](https://github.com/nearprotocol/nearcore/pull/1827) is not merged yet?

looks like we postponed it for now

but generally we need to first flip the switch on staging, make sure stuff works and then push new versions everywhere

PR #202, opened 2019-11-23
ux enhancements re: login experience

amgando:
handful of changes made as independently pickable commits (at least i hope so  🙏)

final result includes:
- quieter output unless `NEAR_DEBUG` is truthy
- more prominent message styling
- web browser opens to auth URL automatically ✨ 
- 3 error cases handled:
  - invalid account (fails by matching `/Account ID/` in error message)
  - account not found on current network
  - public key mismatch for account ID

_(note: some minor formatting fixes were applied after recording the video demo and again after taking the screenshots below.  font size difference in screenshots is bc of image width here)_

30 second demo includes 3 cases: invalid, not found, success 
see here: https://asciinema.org/a/02smCQ2fOhIR0B6xICJegsFyC

**invalid account**
image

**account not found**
image

**public key mismatch**
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,12 +65,15 @@ exports.login = async function(options) {
                 let account = await near.account(accountId);
                 let keys = await account.getAccessKeys();
                 let publicKey = keyPair.getPublicKey().toString();
+                const short = (key) => key.substring(0,14); // keep the public key readable
+
                 let keyFound = keys.some(key => key.public_key == keyPair.getPublicKey().toString());
                 if (keyFound) {
                     await options.keyStore.setKey(options.networkId, accountId, keyPair);
                     console.log(`Logged in as ${accountId} with public key ${publicKey} successfully`);
+                    console.log(chalk`Logged in as [ {bold ${accountId}} ] with public key [ {bold ${short(publicKey)}} ] successfully`);
                 } else {
-                    console.log('Log in did not succeed. Please try again.');
+                    console.log(chalk`The account you provided has not {bold.red authorized the expected key:} [ {bold ${short(publicKey)}... ]} Please try again.\n`);
                 }
             } catch (e) {
                 console.log(e);
--- a/index.js
+++ b/index.js
@@ -72,7 +72,7 @@ exports.login = async function(options) {
                 let account = await near.account(accountId);
                 let keys = await account.getAccessKeys();
                 let publicKey = keyPair.getPublicKey().toString();
-                const short = (key) => key.substring(0,14); // keep the public key readable
+                const short = (key) => `${key.substring(0,14)}...`; // keep the public key readable
 
                 let keyFound = keys.some(key => key.public_key == keyPair.getPublicKey().toString());
                 if (keyFound) {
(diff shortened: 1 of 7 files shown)


vgrichina:
@amgando do you plan to look into CI failure?

index.js
```diff
@@ -80,8 +82,13 @@ exports.login = async function(options) {
         newUrl.searchParams.set('title', title);
         const keyPair = await KeyPair.fromRandom('ed25519');
         newUrl.searchParams.set('public_key', keyPair.getPublicKey());
-        console.log(chalk`\n(Step 1) {bold.yellow Please navigate to this url} and follow the instructions to log in: \n${newUrl.toString()}`);
+        console.log(chalk`\n(Step 1) {bold.yellow Please authorize NEAR Shell} on at least one of your accounts then come back.`);
 
+        if(isCI || process.env.NEAR_DEBUG) {
```
vgrichina: Let's log URL in any case, opening URL can easily fail.

index.js
```diff
@@ -6,6 +6,8 @@ const rimraf = require('rimraf');
 const readline = require('readline');
 const URL = require('url').URL;
 const chalk = require('chalk');  // colorize output
+const open = require('open');    // open URL in default browser
+const isCI = require('is-ci');   // honor CI server limitations
```
vgrichina: let's not use `is-ci` package / have CI-specific code

vgrichina:
Thanks for improvements, please address comments though.

BTW, something that also can be useful – automatically callback from wallet to shell, here's approximate implementation:
https://gist.github.com/vgrichina/cfa393ecb3340779bf42f5a0c02d592f 

In that case we'll need manual account ID prompt only as backup.

index.js
amgando: won't trying to open a browser on a CI server cause problems though?

vgrichina: it likely will just fail and you need to catch error or smth. What happens locally if you pipe stdout to file for example?

amgando:
> Thanks for improvements, please address comments though.

sure thing, will try to make time before next week to address.  clear direction in comments, thx!

> BTW, something that also can be useful – automatically callback from wallet to shell, here's approximate implementation:
> https://gist.github.com/vgrichina/cfa393ecb3340779bf42f5a0c02d592f
> 
> In that case we'll need manual account ID prompt only as backup.

won't this likely prompt the user to allow the open connection?  and what about proxy stuff?  this is a little out of my comfort zone, dealing with all the edge cases of opening a local server as a side effect of a CLI command

amgando:
> @amgando do you plan to look into CI failure?

the CI failures i've seen from my PRs are coming from 2 things: 
- forgetting to lint 🤦‍♂ 
- `masterAccount` parameter missing (since recent updates to NEAR Shell actually)

vgrichina:
> won't this likely prompt the user to allow the open connection? and what about proxy stuff? this is a little out of my comfort zone, dealing with all the edge cases of opening a local server as a side effect of a CLI command

usually you don't need anything special to open connection on localhost (it's doesn't need to be externally facing)

amgando:
oh, of course, it's just a browser redirect. that makes sense

ok, I think I have everything I need to clean up this PR

amgando:
tests are still failing due to missing `masterAccount` param

```
Missing required argument: masterAccount
./test/test_contract.sh FAIL
```

otherwise, @vgrichina, how close is this to ready?

PR #226, opened 2019-12-23
Fixed the update instructions

frol:
See https://github.com/nearprotocol/stakewars/issues/103


frol:
CI failure is unrelated to the change.

PR #227, opened 2019-12-23
Convert --initialBalance from NEAR tokens to integer units 

vgrichina:


commands/create-account.js
```diff
@@ -0,0 +1,50 @@
...
+        })
+        .option('publicKey', {
+            desc: 'Public key to initialize the account with',
+            type: 'string',
+            required: false
+        })
+        .option('initialBalance', {
+            desc: 'Number of tokens to transfer to newly created account',
+            type: 'string',
+            default: '0.1'
```
icerove: why we change it to 0.1?

vgrichina: So that if you have created your master account in wallet, you can create multiple accounts in shell without completely draining master account balance (it's 10 NEAR by default).

commands/create-account.js
```diff
@@ -0,0 +1,50 @@
...
+    let near = await connect(options);
+    let keyPair;
+    let publicKey;
+    if (options.publicKey) {
+        publicKey = options.publicKey;
+    } else {
+        keyPair = await KeyPair.fromRandom('ed25519');
+        publicKey = keyPair.getPublicKey();
+    }
+    await near.createAccount(options.accountId, publicKey);
```
icerove: I wanna know the function createAccount here is never received the arg initialBalance, how could it really pass the commend arg initialBalance to create new account with specific token?

vgrichina: see above:
```
    // NOTE: initialBalance is passed as part of config here
    let near = await connect(options);
```

commands/create-account.js
icerove: I think it is not a good way to do so. give more instructions to user what `create_account` command is from and how to use it make it more clear.

vgrichina: do you suggest not having default amount at all? or different default amount?

commands/create-account.js
icerove: got it

commands/create-account.js
icerove: maybe not having default amount but required amount from user is better. since 0.1 or even less is not good number to guess how many account user want.

commands/create-account.js
```diff
@@ -0,0 +1,50 @@
+
+const exitOnError = require('../utils/exit-on-error');
+const connect = require('../utils/connect');
+const { KeyPair, utils } = require('nearlib');
+
+module.exports = {
```
janedegtiareva: +1 to pulling out commands into separate files!

commands/create-account.js
vgrichina: > maybe not having default amount but required amount from user is better.

I agree there is some trade off between discoverability and convenience here. However I highly prefer tools where you can succeed most of the time with reasonable details. We need to make sure to have `--initialBalance` example in docs though /cc @amgando

PR #231, opened 2020-01-03
Added engine strict to have better ux for old node versions

janedegtiareva:


janedegtiareva:
My understanding is that this would prevent npm install -g near-shell if the node version is too low.

package.json
```diff
@@ -5,6 +5,7 @@
   "engines": {
     "node": ">= 10"
   },
+  "engineStrict": true,
```
vgrichina: Looks like it's been renamed to `engine-strict` quite some time ago https://docs.npmjs.com/files/package.json#enginestrict

Please check if it actually works by installing older Node version e.g. using https://github.com/tj/n

janedegtiareva: does preview version work for npm install -g too?

vgrichina: I haven't tried, but I assume it'll support all normal package sources (including local path, Git repo, etc)

janedegtiareva: adding this option doesn't do much for npm install -g (it still succeeds on inappropriate node version), but also doesn't break anything. Do we want to keep this or scrap? I have no preference at this point

vgrichina: probably no sense to keep as is.

but we can change `bin/near` to use Node as well instead of shell script and do the check similar to how https://github.com/dawsbot/engines-enforcer does it

PR #232, opened 2020-01-06
Document walletUrl option (it already works, just doesn't show in help)

janedegtiareva:


janedegtiareva:
related to https://github.com/nearprotocol/near-shell/issues/111

PR #240, opened 2020-01-14
Prototype dev-deploy

vgrichina:
Add `dev-deploy` command for throwaway dev deploys

### DIFF
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -1,0 +1,66 @@
+const { KeyPair } = require('nearlib');
+const exitOnError = require('../utils/exit-on-error');
+const connect = require('../utils/connect');
+const { readFile, writeFile } = require('fs').promises;
+
+module.exports = {
+    command: 'dev-deploy <wasmFile>',
+    desc: 'deploy your smart contract using temporary account (TestNet only)',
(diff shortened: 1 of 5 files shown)


potatodepaulo:
Rad feature!

commands/dev-deploy.js
```diff
@@ -0,0 +1,66 @@
...
+    console.log(
+        `Starting deployment. Account id: ${accountId}, node: ${nodeUrl}, helper: ${helperUrl}, file: ${wasmFile}`);
+    const contractData = [await readFile(wasmFile)];
+    const account = await near.account(accountId);
+    await account.deployContract(contractData);
+    console.log(`Done deploying to ${accountId}`);
+}
+
+async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
+    const accountFilePath = `${keyStore.keyDir}/dev-account`;
```
potatodepaulo: Long term, we should still shoot for global file path. This implementation doesn't exclude that.

vgrichina: We need to think more on this. I feel that stuff like current dev account likely belongs to project folder (as is project specific), but production keys should better go to some more secure place (e.g. OS X key chain).

PR #244, opened 2020-01-16
Use flagged-respawn to respawn Node.js cleanly for `near repl`

vgrichina:
This also enables user to pass desired Node.js/V8 flags


vgrichina:
Have to merge this as without the fix `near` fails to work with latest `yarn`

PR #250, opened 2020-01-24
cli tells user keys are in ./neardev instead of ~/.near

mikedotexe:
As I understand, `near-core` places validator keys in `~/.near` but shell looks for keys in the project's `./neardev` directory.
https://github.com/nearprotocol/near-shell/blob/master/middleware/key-store.js#L7

With this PR, if a user types `near --help` they will see cli messages explaining this.

Screenshot 2020-01-23 20 00 18

Smallest pull request in my life. 🎊

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,6 @@ near <command>
 | --networkId               | NEAR network ID for different keys by network | [string]  |"default"              |
 | --helperUrl               | NEAR contract helper URL                      | [string]  |                       |
 | --keyPath                 | Path to master account key                    | [string]  |                       |
-| --homeDir                 | Where to look for master account              | [string]  |"~/.near"              |
 | --accountId               | Unique identifier for the account             | [string]  [required]|             |
 | --masterAccount           | Account used to create requested account.     | [string]  [required]|             |
 | --publicKey               | Public key to initialize the account with     | [string]  [required]|             |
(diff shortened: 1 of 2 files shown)


vgrichina:
@mikedotexe I think `--homeDir` argument isn't really used anywhere, so need to either remove it and fix the docs or actually implement it.

mikedotexe:
You're right, Vlad, thanks.
I think this would be a good ticket for the icebox when we implement Max's direction for using ZenHub after the offsite.
Keeping track of related issues here:
https://github.com/nearprotocol/docs/issues/155
appears in `./docs/development/near-clitool.md`
https://github.com/nearprotocol/near-shell/issues/159

vgrichina:
@mikedotexe icebox is basically stuff which never gets done. It's best to fix small issues like these right away instead, or it will keep getting bounced forever.

mikedotexe:
> @mikedotexe icebox is basically stuff which never gets done. It's best to fix small issues like these right away instead, or it will keep getting bounced forever.

Okay. We've had a discussion where Damon is going to create an NEP for what we'd like to see in near shell in the future. I feel like at this point in time it makes more sense to remove that argument and move on to pressing issues. This won't be hard, but I'll get this in soon so we can close this issue and have slightly less confusion for the end user.

Issue #258, opened 2020-02-01
Inconsistent output about NEAR on function call


ilblackdragon:
Calling function with 10 N attached, outputs `10000000000000000000000000 NEAR`.

```
near call my_validator deposit '' --accountId acc1 --amount 10
[WARNING] Didn't find config at /Users/cypress/projects/nearprotocol/stakewars/staking-contract/src/config, using default shell config
{ networkId: 'local',
  nodeUrl: 'http://localhost:3030',
  keyPath: '/Users/cypress/.near/validator_key.json',
  walletUrl: 'http://localhost:4000/wallet',
  contractName: 'near-hello-devnet' }
(node:53150) ExperimentalWarning: The fs.promises API is experimental
Scheduling a call: my_validator.deposit() with attached 10000000000000000000000000 NEAR
```

PR #261, opened 2020-02-03
Clarity on yoctonear #258

mikedotexe:
Address issue seen in #258 
Essentially we were calling nearlib's `parseNearAmount` and showing the user that value as if it were NEAR, when in truth it's the yoctoNEAR denomination.
I saw this on `call` and `send` commands.
Now:
Screenshot 2020-02-03 10 09 31
Screenshot 2020-02-03 10 10 18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 tmp-project
 neardev/*
 package-lock.json
+.idea
--- a/commands/call.js
+++ b/commands/call.js
@@ -23,7 +23,7 @@ module.exports = {
 
 async function scheduleFunctionCall(options) {
     console.log(`Scheduling a call: ${options.contractName}.${options.methodName}(${options.args || ''})` +
-        (options.amount && options.amount != '0' ? ` with attached ${utils.format.parseNearAmount(options.amount)} NEAR` : ''));
+        (options.amount && options.amount != '0' ? ` with attached ${options.amount} NEAR (${utils.format.parseNearAmount(options.amount)} yoctoNEAR)` : ''));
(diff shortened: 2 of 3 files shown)


janedegtiareva:
Do you think printing yoctoNEAR actually adds value? It's a large number with a lot of 0's and is usually not very human readable.

mikedotexe:
> Do you think printing yoctoNEAR actually adds value? It's a large number with a lot of 0's and is usually not very human readable.

The reason why I kept that in was in case they're running `near tx-status` or otherwise seeing gas values come back that seem very large. This primes an end-user's brain to recognize large values if they look into the details of a transaction. Happy to remove it if we think it's too much information.

chadoh:
Can we show NEAR with scientific notation, or appropriate metric prefixes, instead of giant yoctoNEAR values? I would rather see "10 nⓃ" or "1e-8 Ⓝ" rather than "10,000,000,000,000,000 yoctoⓃ"

mikedotexe:
Alright, I've removed the yoctonear verbiage, sounds like that's the consensus. This will reduce the confusion addressed in the initial issue. Read for a quick re-review, friends

vgrichina:
@mikedotexe can this be merged?

mikedotexe:
> @mikedotexe can this be merged?

definitely. I'm still getting used to the idea of merging my own PR. I come from a place where that was forbidden. Seems like I should have done that here, though.

vgrichina:
@mikedotexe yeah, if it's reviewed and you made sure it works well – feel free to just merge.

Issue #259, opened 2020-02-01
Invalid subcommand doesn't show any errors


ilblackdragon:
Currently running non-existent / invalid sub-command doesn't result in any error.
```
near transfer acc1 my_validator 1
[WARNING] Didn't find config at /Users/cypress/projects/nearprotocol/stakewars/staking-contract/src/config, using default shell config
{ networkId: 'local',
  nodeUrl: 'http://localhost:3030',
  keyPath: '/Users/cypress/.near/validator_key.json',
  walletUrl: 'http://localhost:4000/wallet',
  contractName: 'near-hello-devnet' }
(node:53295) ExperimentalWarning: The fs.promises API is experimental
```

vgrichina:
@mikedotexe related PR https://github.com/nearprotocol/near-shell/pull/51

Note that just re-enabling strict mode is problematic, as then loading config fails when command doesn't support all of the options in config.

It should be possible to solve by loading and merging config with options manually vs going through yargs. Best place to put it is likely yargs middleware:
https://github.com/yargs/yargs/blob/master/docs/advanced.md#middleware

mikedotexe:
> @mikedotexe related PR #51
> 
> Note that just re-enabling strict mode is problematic, as then loading config fails when command doesn't support all of the options in config.
> 
> It should be possible to solve by loading and merging config with options manually vs going through yargs. Best place to put it is likely yargs middleware:
> https://github.com/yargs/yargs/blob/master/docs/advanced.md#middleware

Super helpful context, thank you. I took another approach here
https://github.com/nearprotocol/near-shell/pull/262
that doesn't use the strict mode, but has a default.
Knowing the history of this issue, I'll tag Jane and you on it since you'll immediately understand

mikedotexe:
Finally got this through, thanks for the patience, Illia.

PR #262, opened 2020-02-03
Add Yargs default and middleware logic to display helpful message during invalid arg/command calls

mikedotexe:
Regarding #259 
[Followed this](https://github.com/yargs/yargs/blob/master/docs/advanced.md#default-commands) from the yargs documentation.

Added a couple of cases where an argument, command, and multiple commands are entered an not understood. Provides end user with simple feedback and instructions.

Screenshot 2020-02-03 11 52 01

**Update**: please see comments below regarding evolved approach

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -134,9 +134,31 @@ const deploy = {
 registerCommand(deploy);
 
 const callViewFunction = {
-    command: 'view <contractName> <methodName> [args]',
+    command: 'view [contractName] [methodName] [data] [gas] [amount]',
     desc: 'make smart contract call which can view state',
-    builder: (yargs) => yargs,
+    builder: (yargs) => yargs
+        .option('contractName', {
+            desc: 'Contract where the method is being called',
+            type: 'string',
+        })    
+        .option('methodName', {
+            desc: 'Contract method being called',
+            type: 'string',
+        })
+        .option('data', {
+            desc: 'Data to send to method',
+            type: 'string',
+        })
+        .option('gas', {
+            desc: 'Max amount of gas this call can use',
+            type: 'string',
+            default: '100000000000000'
+        })
+        .option('amount', {
+            desc: 'Number of tokens to attach',
+            type: 'string',
+            default: '0'
+        }),        
     handler: exitOnError(main.callViewFunction)
 };
 registerCommand(callViewFunction);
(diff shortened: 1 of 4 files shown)


vgrichina:
@mikedotexe see here https://github.com/nearprotocol/near-shell/issues/259#issuecomment-581564218, I think we should do it differently, so that arguments are also validated.

However feel free to merge it as a quick fix for now if full fix requires much more time.

mikedotexe:
Hey folks, I have prefixed this PR with "WIP:" meaning work-in-progress. It's in a good state to start reviewing, I will make some minor changes.
So I am agreed that `strict()` will not work for us unless we change the config, which is possible but not ideal.
@vgrichina has the right idea of using middleware. A strange combination happens when there's middleware and a default command, both of which we need. The problem is that running `near` without arguments or commands no longer shows the full command list like it was when running `near --help`
This is [a pesky issue](https://github.com/yargs/yargs/issues/895).
I did not want to sacrifice this behavior, so I ripped out and modified some logic so that we'll still show all the commands and descriptions.
I also chained `.fail()` onto this with some minor error logging and suggestions.

Result:
`near` and `near --help` report the same screen (tiny difference is the word "near" appears in a different spot)
`near fake` » `Invalid command 'fake'`
`near fake --fakeArg hi` » `Invalid command 'fake'` `Invalid argument 'fakeArg`
`near call fake` » `Error: Not enough non-option arguments: got 1, need at least 2` `Please use the --help flag for more information.`
`near login fake` » `Invalid command 'fake'`

vgrichina:
@mikedotexe the issue you linked seems to have a solution:

https://github.com/yargs/yargs/issues/895#issuecomment-392893305

any reason why we cannot use it? 

if there is bug in yargs, I'd prefer we fix it vs reimplementing half of the logic ourselves

mikedotexe:
> @mikedotexe the issue you linked seems to have a solution:
> 
> [yargs/yargs#895 (comment)](https://github.com/yargs/yargs/issues/895#issuecomment-392893305)
> 
> any reason why we cannot use it?
> 
> if there is bug in yargs, I'd prefer we fix it vs reimplementing half of the logic ourselves

The comment you mentioned uses `strict()` which, as I understand will always throw for `contractName` and `walletUrl` that exists in `config.js`
I also don't like the idea of borrowing 54 lines of code from yargs, but having the commands there made it significantly better in my opinion. This is one of those cases where I was fighting the framework.

I think I'm now understanding Jane's comment better and will look into the approach.

ilblackdragon:
What is the status with this PR?

mikedotexe:
This approach becomes difficult with `near call` and `near view`. Closing in favor of https://github.com/nearprotocol/near-shell/pull/265

PR #265, opened 2020-02-06
add yarg’s strict() to chain. add walletUrl, contractName as options

mikedotexe:
Different approach to https://github.com/nearprotocol/near-shell/issues/259
Also enabled recommendations. So `near fake` will fail, list possible commands, and say, "Did you mean `near stake`?"
I think the issue was `.config(config)` loaded `src/config.js` which has keys that weren't explicitly set as `option`s.

~This is quite separate from the PR where most of the discussion on this issue has been happening.~
https://github.com/nearprotocol/near-shell/pull/262

If we choose to merge this PR, we can close the other.

Update: this PR has been languishing and Illia brought up that this is important. I see that @vgrichina suggests we overhaul how config is used. Can we get something done in the meantime?

Tests have been failing and I have commits fixing these, including some outdated shell commands. Kind of mysterious that tests have worked at all. I guess some errant calls technically don't `exit(1)`? Doesn't matter too much, this is ready for re-review.

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -183,7 +183,7 @@ yargs // eslint-disable-line
         hidden: true
     })
     .option('masterAccount', {
-        desc: 'Master account used when create new accounts',
+        desc: 'Master account used when creating new accounts',
         type: 'string',
         hidden: true
     })
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -173,11 +173,12 @@ yargs // eslint-disable-line
     .option('walletUrl', {
         desc: 'Website for NEAR Wallet',
         type: 'string',
-        default: 'trythis'
+        hidden: true
     })
     .option('contractName', {
         desc: 'Account name of contract',
         type: 'string',
+        hidden: true      
     })
     .middleware(require('../middleware/key-store'))
     .command(require('../commands/create-account'))
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -182,6 +182,11 @@ yargs // eslint-disable-line
         type: 'string',
         hidden: true
     })
+    .option('masterAccount', {
+      desc: 'Master account used when create new accounts',
+      type: 'string',
+      hidden: true
+    })
     .middleware(require('../middleware/print-options'))
     .middleware(require('../middleware/key-store'))
     .command(require('../commands/create-account'))
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -183,9 +183,9 @@ yargs // eslint-disable-line
         hidden: true
     })
     .option('masterAccount', {
-      desc: 'Master account used when create new accounts',
-      type: 'string',
-      hidden: true
+        desc: 'Master account used when create new accounts',
+        type: 'string',
+        hidden: true
     })
     .middleware(require('../middleware/print-options'))
     .middleware(require('../middleware/key-store'))
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -145,6 +145,7 @@ const clean = {
 
 let config = require('../get-config')();
 yargs // eslint-disable-line
+    .strict()
     .middleware(require('../utils/check-version'))
     .scriptName('near')
     .option('nodeUrl', {
@@ -167,6 +168,15 @@ yargs // eslint-disable-line
     })
     .option('accountId', {
         desc: 'Unique identifier for the account',
+        type: 'string',
+    })
+    .option('walletUrl', {
+        desc: 'Website for NEAR Wallet',
+        type: 'string',
+        default: 'trythis'
+    })
+    .option('contractName', {
+        desc: 'Account name of contract',
         type: 'string',
     })
     .middleware(require('../middleware/key-store'))
@@ -197,5 +207,6 @@ yargs // eslint-disable-line
     })
     .showHelpOnFail(true)
     .demandCommand(1, 'Please enter a command')
+    .recommendCommands()
     .wrap(null)
     .argv;
--- a/test/test_account_operations.sh
+++ b/test/test_account_operations.sh
@@ -17,4 +17,4 @@ if [[ ! "$RESULT" =~ $EXPECTED ]]; then
     exit 1
 fi
 
-../bin/near delete_account $testaccount test.near
+../bin/near delete $testaccount test.near
--- a/test/test_generate_key.sh
+++ b/test/test_generate_key.sh
@@ -1,10 +1,10 @@
 
 #!/bin/bash
 set -ex
-KEY_FILE=neardev/ci/generate-key-test.json
+KEY_FILE=neardev/$NODE_ENV/generate-key-test.json
 rm -f "$KEY_FILE"
 
-RESULT=$(./bin/near generate-key generate-key-test --networkId ci)
+RESULT=$(./bin/near generate-key generate-key-test --networkId $NODE_ENV)
 echo $RESULT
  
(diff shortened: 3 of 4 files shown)


bin/near-cli.js
```diff
@@ -169,6 +170,15 @@ yargs // eslint-disable-line
         desc: 'Unique identifier for the account',
         type: 'string',
     })
+    .option('walletUrl', {
+        desc: 'Website for NEAR Wallet',
+        type: 'string',
+        default: 'trythis'
```
vgrichina: why this default?

vgrichina:
> I think the issue was .config(config) loaded src/config.js which has keys that weren't explicitly set as options.

yes, that was the problem

not sure if it's not confusing to keep these as options for all commands (as most commands would ignore `--contractName` for example) or also what may happen given that it's alias to `--accountId` – we might end up using value from `src/config.js` for commands which don't operate on contracts.

for this reason I feel that long-term we need to figure out solution where we load config in different way

vgrichina:
@mikedotexe this is failing the tests BTW

mikedotexe:
> @mikedotexe this is failing the tests BTW

Yes, I haven't seen any PRs pass tests the past two days.
```
Unknown argument: masterAccount
++RESULT='{ networkId: '\''shared-test'\'',
  nodeUrl: '\''http://shared-test.nearprotocol.com:3030'\'',
  contractName: undefined,
  masterAccount: '\''test.near'\'' }'
./test/test_generate_key.sh FAIL
error Command failed with exit code 1.
```

`masterAccount` is not something I've touched, which is odd.
Locally, when I run `rm -rf neardev && yarn test` on `master` it's also failing, but for a different reason.
```
Account has existing key pair with ed25519:EsyKEPeLbgPk9qorxLZ2thLghLXJXdfMubF8muy5mg2j public key =~ .*Account has existing key pair with ed25519:.+ public key.* ]]
./test/test_generate_key.sh SUCCESS
error Command failed with exit code 1.
```
I've had to somewhat ignore errors as it seems something is awry.

mikedotexe:
The latest commit reduces that confusion. When a user types `near` or `near --help` it does not show the `walletUrl` or `contractName` flags.
I hear what you're saying, @vgrichina, about loading the config. I think this approach will work. I also sense there are going to be a significant changes to shell that are coming soonish. This is a fairly quick approach. I feel we want to improve UX the original issue soon, giving users feedback on what went wrong when they type something incorrect.

bin/near-cli.js
```diff
@@ -170,6 +171,22 @@ yargs // eslint-disable-line
...
+        type: 'string',
+        hidden: true
+    })
+    .option('contractName', {
+        desc: 'Account name of contract',
+        type: 'string',
+        hidden: true
+    })
+    .option('masterAccount', {
+        desc: 'Master account used when create new accounts',
```
chadoh: Please change "used when create" to "used to create" or "used when creating"

bin/near-cli.js
```diff
@@ -170,6 +171,22 @@ yargs // eslint-disable-line
...
+        type: 'string',
+        hidden: true
```
chadoh: Why hide all of these?

test/test_account_operations.sh
```diff
@@ -17,4 +17,4 @@ if [[ ! "$RESULT" =~ $EXPECTED ]]; then
     exit 1
 fi
 
-../bin/near delete_account $testaccount test.near
+../bin/near delete $testaccount test.near
```
chadoh: Is this related to the `yargs` changes? Should it be submitted as a separate PR?

test/test_generate_key.sh
```diff
@@ -18,7 +18,7 @@ if [[ ! "$RESULT" =~ $EXPECTED ]]; then
     exit 1
 fi
 
-RESULT2=$(./bin/near generate-key generate-key-test --networkId default)
+RESULT2=$(./bin/near generate-key generate-key-test --networkId ci)
```
chadoh: Same question about all these changes from `default` to `ci` – should they be in a separate PR? I don't see any description about these.

test/test_account_operations.sh
mikedotexe: never updated from:
https://github.com/nearprotocol/near-shell/commit/434532f9e28ce38ebe8b209b271e1d107b4fdaa3#diff-22d5d2b57c69d2e1ac054c4d0f744ff1L53-R111

test/test_generate_key.sh
mikedotexe: without a modification like this tests break. I don't understand how they haven't been breaking before this. I pushed a change adding an environment variable instead.
it sounds like you're confused and expected more info. how should i communicate this? commit is too short to explain. should I be updating the PR description with each change? At this point I'm firefighting the broken tests. I can add that

bin/near-cli.js
mikedotexe: done

bin/near-cli.js
mikedotexe: This is a good debate to be had. The short answer is, there's a long road to getting `near-shell` consistent, and this is something of a bandaid. Actually, this whole PR is a bandaid.
Note that when you run: `near` to display the default help from yargs, it spits out commands and then the `Options:` section. These are flags that **typically** (again, we're working on it) apply to the top-level commands. Actually, the only good example of this is `--accountId` here.
Then, when you run `near create_account --help` you'll see the relevant `Options:` for that particular command, including `--masterAccount`.
This `hidden` key is not hiding it everywhere, but hiding it from the top-level list. I'm not against removing the `hidden` part, but it'll just make th

mikedotexe: whoops. It'll just make that a bit longer. The fundamental issue here is we want to use `strict` mode, but we can't do that when we load `config.js` and it contains additional flags that aren't captured at the top level. So in this PR we're like, "yeah, you **can** give me a `masterAccount` flag here, it won't do much on its own, but you can"
And from there `yargs`'s `strict` mode is possible whereby it'll scream at you if you misspell something or provide suggestions (`recommendCommands`) or tell you what you're missing.

chadoh: Excellent explanation, thank you

test/test_generate_key.sh
chadoh: Ideally I would like to see one or two separate PRs laser-focused on fixing these issues, as it leaves more of a paper-trail and makes space for better discussion. But whether a totally new PR is necessary is a judgement call, and I'm ok with fitting these small fixes into this one as separate commits, each with a good description.

test/test_account_operations.sh
chadoh: Ok, so this is a broken thing that just needed to be fixed, and the fix is small enough that it may as well get fixed as part of this larger PR?

mikedotexe:
> Looks good! Have you tested it on windows? Do you mind commenting on the PR with how you tested it? LGTM after making sure it works on diff platforms

On Windows I run `near fake` and nothing happens. Then when I'm on this branch I can run:
`node bin\near fake` and it'll suggest a similar, valid command.
2020-03-31 14_39_19-Window

Same thing on OS X, except this time I'll use an errant argument.
`near state --fake argie mikedotdmg`
makes it look like everything is fine, leading the user to believe they may need this `--fake` argument.
running on this PR:
`./bin/near state --fake argie mikedotdmg`
It tells the user that `fake` isn't a thing.

Screenshot 2020-03-31 14 47 22

The first line of this PR's description links to the issue created by Illia who was trying to run:
`near transfer acc1 my_validator 1`

This has the same behavior.

PR #273, opened 2020-02-21
add code of conduct, contributing, license

mikedotexe:
Addresses: https://github.com/nearprotocol/near-shell/issues/267
This is a request that came in a while ago, getting to it with the repos in closest reach.

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/273"></a>


vgrichina:
@mikedotexe we use MIT LICENSE by default I think, nearcore is using GPL only because it historically has used GPL code

vgrichina:
@mikedotexe make sure LICENSE is fixed in other places as well, e.g. I didn't notice https://github.com/nearprotocol/near-wallet/blob/master/LICENSE

mikedotexe:
> @mikedotexe we use MIT LICENSE by default I think, nearcore is using GPL only because it historically has used GPL code

I took the license from `nearcore` here:
https://github.com/nearprotocol/nearcore/blob/master/LICENSE

vgrichina:
@mikedotexe 
> I took the license from nearcore here:

yes, that's what I figured:
> nearcore is using GPL only because it historically has used GPL code

Stuff outside of nearcore shouldn't use GPL, but should use more permissive MIT license (see e.g. https://github.com/nearprotocol/nearlib/blob/master/LICENSE)
AFAIK nearcore also doesn't plan to keep using GPL forever.

Also we promise MIT license in https://github.com/nearprotocol/near-shell/blob/ffeb8db08f570e449610f02078e997be56f5cbc6/package.json#L19, so having different one in repository would be super confusing.

frol:
FYI https://github.com/nearprotocol/near-explorer/pull/177

/cc @AnaisUrlichs

potatodepaulo:
Looks good!

PR #274, opened 2020-02-28
Print config after merging with options

vgrichina:
Fixes https://github.com/nearprotocol/near-shell/issues/243


icerove:
also see the CI tests

vgrichina:
> also see the CI tests

@icerove tests are broken for near-shell `master`, I'm debugging this + also going to execute near-shell tests in create-near-app CI to prevent this from happening

PR #275, opened 2020-02-28
gulp on windows needs shell key during spawn()

mikedotexe:
In testing examples on a Windows machine, I noticed that `near build` will not work.

`Error: spawn gulp ENOENT`

I eventually found this [StackOverflow answer](https://stackoverflow.com/a/54515183). Tested on Windows and Mac.

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -118,7 +118,7 @@ const build = {
     command: 'build',
     desc: 'build your smart contract',
     handler: () => {
-        const gulp = spawn('gulp');
+        const gulp = spawn('gulp', [], {shell: process.platform == 'win32'});
         gulp.stdout.on('data', function (data) {
             console.log(data.toString());
         });


vgrichina:
@mikedotexe let's just remove `near build`. It doesn't make sense anymore given we also have Rust with completely different build system. So if any examples are running `near build` – just update them to do `gulp` directly.

willemneal:
Was talking with Yifang today and we could make it work with both rust and as. Currently projects in As should have a asconfig.js, and the rust projects have a build.sh file.

icerove:
@vgrichina  have a look at what Willem said and I agree with him. Still want your opinion.

vgrichina:
@willemneal @icerove  what is the motivation to keep `near build` though? It doesn't seem to be used by anyone (this seems like only place in docs mentioning it https://github.com/nearprotocol/docs/blob/fe6f20e432e93c7b93640ec01e3901afcd4884de/docs/development/near-clitool.md#L102).

mikedotexe:
> @willemneal @icerove what is the motivation to keep `near build` though? It doesn't seem to be used by anyone (this seems like only place in docs mentioning it https://github.com/nearprotocol/docs/blob/fe6f20e432e93c7b93640ec01e3901afcd4884de/docs/development/near-clitool.md#L102).

From a user experience (also DevX) I think it would be a good idea to keep it.
Then NEAR Shell is the one-stop shop. 
`near build` then `near deploy` and `near call`
If a user gets confused, they hit `near --help` and see what they can do.

vgrichina:
> Then NEAR Shell is the one-stop shop.

it isn't really. Like you need to do `cargo test` to test your Rust smart contracts for example. I prefer we expose platform-specific build tools to avoid extra layer of confusion.

willemneal:
@vgrichina No one was using `near build` because all our examples and repo used `gulp` and now `node asconfig.js`.  I agree with platform specific tools, but near cli should be able to detect their presence and invoke them.  And we can still have docs about them.

Heck ideally we could even have near-shell handle the installation step.  And then every `near` project repo can be cloned and then run with near-shell.

mikedotexe:
> @vgrichina No one was using `near build` because all our examples and repo used `gulp` and now `node asconfig.js`. I agree with platform specific tools, but near cli should be able to detect their presence and invoke them. And we can still have docs about them.
> 
> Heck ideally we could even have near-shell handle the installation step. And then every `near` project repo can be cloned and then run with near-shell.

`near build` just calls `gulp`, which worked for the AS examples

vgrichina:
> Heck ideally we could even have near-shell handle the installation step. And then every near project repo can be cloned and then run with near-shell.

Why do you think it's better then just using platform-specific tools directly? Truffle (Ethereum dev stuff) is wrapping around tools like mocha, etc and it is total pain in the ass to work with. With using `mocha` directly I can for example filter tests by pattern. But in their wrapper I cannot do it. We are very likely to run into similar limitations if following same approach. 

It's going to become even worse when we start doing any of these:
1) use languages other than Rust and AssemblyScript (do we now have to support every third-party tooling in `near build`?)
2) have multi-contract projects with custom builds. If it's `gulp` or `cargo` for building – I can just use whatever tools platform provides to customize my build. If it's `near build` – I need first to figure out how it even works. 

BTW, I don't think `node asconfig.js` is good idea either – many projects will have more than 1 smart contract. We need to think of solution which can scale for more complicated build than 1 contract with no dependencies.

potatodepaulo:
Seems like this discussion on this PR has blown out of scope. I took a look and it's a 1-line change to fix build on windows. I propose we merge this to patch the shell, and then move the philosophical discussion to the design of the new shell: https://github.com/nearprotocol/NEPs/pull/31
Also discussion here can be revived: https://commonwealth.im/near/proposal/discussion/234-near-shell-ideations

PR #282, opened 2020-03-10
fix: correct typo in error message "netiher"

chadoh:


mikedotexe:
I just came here to do this, nice work.

PR #298, opened 2020-03-31
Set initialBalance to 100N in create_account

ilblackdragon:
After state staking update https://github.com/nearprotocol/NEPs/pull/41, 1N covers for 10kb of storage. 
So 100N covers 1MB storage which should be enough for initial playing with contracts.
Current 0.1N leads to LackBalance error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 10
   - 12
 env:
   - NODE_ENV=ci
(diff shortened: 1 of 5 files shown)


potatodepaulo:
I want to highlight that accounts created in the wallet and funding these would need to have 10,000 to keep the same ratio that we've got now.

ilblackdragon:
Why do we need to maintain same ratio?

ilblackdragon:
Hm @mikedotexe it seems like CI has been broken before?

mikedotexe:
> Hm @mikedotexe it seems like CI has been broken before?

Yes, we just fixed that today. Node version and shell tests had minor issues.

mikedotexe:
Actually, there's a test that was checking the balance, too, which has now changed. Pushing up a commit modifying that test.

PR #299, opened 2020-04-01
Fix login prompt when web server redirect doesn’t work

vgrichina:
Fixes https://github.com/nearprotocol/near-shell/issues/297

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,22 +57,21 @@ exports.login = async function(options) {
         newUrl.searchParams.set('public_key', keyPair.getPublicKey());
 
         console.log(chalk`\n{bold.yellow Please authorize NEAR Shell} on at least one of your accounts.`);
-        console.log(chalk`\n{dim If your browser doesn't automatically open, please visit this URL\n${newUrl.toString()}}`);
 
         // attempt to capture accountId automatically via browser callback
         let tempUrl;
-        let accountId;
 
         // find a callback URL on the local machine
         try {
             tempUrl = await capture.callback();
         } catch (error) {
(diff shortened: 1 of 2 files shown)


mikedotexe:
Opened the link on another computer, and after the authorizing in Wallet the last step is to enter you account name. Then it redirected me to:
http://127.0.0.1:3000/?account_id=mike-exe&public_key=ed25519%3A5euofQcDVc497NwYVwoLXZR1c86BWFcNqc2nzBtsxzfq&all_keys=ed25519%3A9Zyt2YKgSaUHNwFzRQxyGHYt6vFXjdLa4nwKQTaBPiiQ

>This site can’t be reached

Which looks pretty broken. This is probably a Wallet issue? I'm not super familiar with this workflow yet, but I'll make an issue in Wallet or shell when I dig in.

janedegtiareva:
nit: let's start leaving some notes on what testing was done in PRs

mikedotexe:
ran `near login` from Computer A, copied link over to Computer B. Computer B brings me to Wallet, I go through the steps, after the very last step I see this:
Screenshot 2020-04-01 09 50 47

because the URL generated by the new login flow adds this param:

    success_url=http%3A%2F%2F127.0.0.1%3A3000

That is a localhost address and port that was determined to be open and available on Computer A.
This is a good stop-gap and I support merging it to unblock validators.
I think the next step is to do what @potatodepaulo suggested and have a prompt for the user. Something like, "Are you logging in from a server with no browser? (default no)"

Then the URL posted would be different, including or excluding where the `success_url` param goes.

PR #300, opened 2020-04-01
Add GitPod support

vgrichina:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/300"></a>

### DIFF
--- a/utils/capture-login-success.js
+++ b/utils/capture-login-success.js
@@ -13,30 +13,35 @@ let server;
     @param port the port the server should use
     @param hostname the hostname the server should use
  */
-const payload = (fields, { port, hostname }) => new Promise((resolve, reject) => {
-
-    const message = renderWebPage('You are logged in. Please close this window.');
+const payload = (fields, { port, hostname }, redirectUrl) => new Promise((resolve, reject) => {
     server = stoppable(http.createServer(handler)).listen(port, hostname);
 
     /**
         request handler for single-use node server
      */
     function handler(req, res){
-        res.statusCode = 200;
-        res.setHeader('Content-Type', 'text/html');
         try {
             let parsedUrl = url.parse(req.url, true);
             let results = fields.map((field) => parsedUrl.query[field]);
-            res.end(message, () => {
-                server.stop();
-                resolve(results);
-            });
+
+            if (Object.keys(parsedUrl.query).length > 0) {
+                res.statusCode = 200;
+                res.setHeader('Content-Type', 'text/html');
+                // TODO: Make code more specialized (vs handling generic fields) and only output this if login succeeded end to end
+                res.end(renderWebPage('You are logged in. Please close this window.'), () => {
+                    server.stop();
+                    resolve(results);
+                });
+            } else {
+                res.writeHead(301, { Location: redirectUrl });
+                res.end();
+            }
         } catch (e) {
-            console.log('Unexpected error: ', e);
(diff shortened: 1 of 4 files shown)


mikedotexe:
Hm, this might have been a glitch with Gitpod, but I ran into an issue.
When I opened this in Gitpod, I went to the Theia Terminal and ran:
`./bin/near login`

A prompt came up asking me how I wanted to open it. There were three options. "Make public, preview, or open in browser" I chose "open in browser"
A new tab opened and I saw this:
Screenshot 2020-04-01 15 04 50

Then I reloaded that page (thinking that Gitpod might have needed a moment to load) and I saw the success message immediately, even though I hadn't logged into an account. 

>You are logged in. Please close this window.

Then back in Gitpod it reads:

>The account you provided [ undefined ] has not authorized the expected key [ ed25519:8bwaTu... ]  Please try again.

mikedotexe:
Yeah something seems to be awry here. I tried stopping the Gitpod workspace instance and restarting it. (Note that when a link tries to open in Gitpod, it'll ask you a prompt and can save your choice for later for certain prompts)
This time I clicked the "make public" option when running `./bin/near login`
And it then gives me this option:

Screenshot 2020-04-01 15 12 13

I tried both of the options and am still getting the same behavior as [the earlier comment](https://github.com/nearprotocol/near-shell/pull/300#issuecomment-607514152).

vgrichina:
@mikedotexe I see, I think the problem is that server gets immediately stopped once you visit the page (no matter whether you got redirected there or opened yourself as in your case).

I think I'll just add smth to GitPod config to avoid this popup (everything works well if you don't click any of these).

mikedotexe:
I'm still seeing that popup asking me to open it in preview or a browser tab. 

Screenshot 2020-04-01 17 31 36

I _can_, however, ignore that and cmd + click on the link in the Terminal and it will work until I re-enter my account name. Then I get:
Screenshot 2020-04-01 17 33 21

Is there a different path I'm supposed to be following here? Seems like the average Gitpod user will be quite confused by the UX presented at this moment.

vgrichina:
@mikedotexe I'm debugging for now, this is definitely not the UX that is supposed to stay

vgrichina:
@mikedotexe try now. Should generally work fine both in popup/preview/blocked popup.

However looks we are hitting some wallet bug where redirect URL is sometimes forgotten, digging more into it.

vgrichina:
> looks we are hitting some wallet bug where redirect URL is sometimes forgotten, digging more into it.

Seems like it's just been this error:

> Error: Transaction HebBhYF4sNUyMdfh3HNDa1b6GXc4Xfi9PXoshj54ZyTb failed. The public key ed25519:2QK2j76EhVwNkAvz1TmnHZYWnngYpKt5L3wM6mUsWEBZ is already used for an existing access key

Tracked here https://github.com/nearprotocol/near-wallet/issues/429

utils/capture-login-success.js
```diff
@@ -13,30 +13,35 @@ let server;
...
+                });
+            } else {
+                res.writeHead(301, { Location: redirectUrl });
+                res.end();
+            }
         } catch (e) {
-            console.log('Unexpected error: ', e);
+            console.error('Unexpected error: ', e);
             res.statusCode = 400;
             res.end('It\'s a scam!');
```
mikedotexe, line 42: whaaaa 😆

PR #308, opened 2020-04-08
Replace ci-staging with ci-betanet

vgrichina:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/308"></a>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ node_js:
   - 12
 env:
   - NODE_ENV=ci
-  - NODE_ENV=ci-staging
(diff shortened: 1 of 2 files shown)


amgando:
@vgrichina heads up, this is causing CI tests to fail in more recent PRs like #341  

not sure what the right move is since the `ci-` prefix was dropped from `config.js`

vgrichina:
> not sure what the right move is since the ci- prefix was dropped from config.js

@amgando what do you mean? Where? How are tests executed then?

PR #313, opened 2020-04-16
Initial stab at dev event tracking (E2E with mixpanel works)

janedegtiareva:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/313"></a>


eventtracking.js
```diff
@@ -0,0 +1,44 @@
...
+    console.log(nearPath)
+    if (!fs.existsSync(nearPath)) { // TODO: make a util for .near
+        //create near path if it doesn't already exist
+        fs.mkdirSync(nearPath);
+    }
+
+    const shellSettingsPath = path.join(nearPath, 'shellSettings');
+    if (!fs.existsSync(shellSettingsPath)) {
+        //Ask if it's ok to collect feedback
+        console.log("We would like to collect data on near shell usage to improve developer experience. Would you like to opt in?")
```
mikedotexe, line 19: I think it would be great to use inquirer here:
https://www.npmjs.com/package/inquirer

I very much plan to use that at some point with `near-shell`

mikedotexe:
I've got a lot of the shell NEP live here:
https://github.com/nearprotocol/NEPs/blob/6b6e9ce2ab5f9cb36b69accae3970d8ea997a29b/text/0000-near-shell.md#upgradability

It would be great if this were stored like this, if possible:
in `~/.near-config/settings.js`:

```
{
  "metricsOptIn": true
}
```

package.json
```diff
@@ -37,13 +37,17 @@
     "flagged-respawn": "^1.0.1",
     "is-ci": "^2.0.0",
     "jest-environment-node": "^25.1.0",
+    "mixpanel": "^0.11.0",
     "ncp": "^2.0.0",
     "near-api-js": "^0.23.1",
     "open": "^7.0.1",
+    "os": "^0.1.1",
```
vgrichina, line 44: why `os` and `path` added as external deps? There should be built-in modules normally.

eventtracking.js
vgrichina: Unless we actually need it's features (selecting from list, etc), it makes sense to just use built-in https://nodejs.org/api/readline.html

At least let's make sure only one of these gets used.

janedegtiareva: I agree with only one way to get input. I am planning to use whatever is in login when I make this production quality (needs a couple of passes)

package.json
janedegtiareva: thanks! Didn't know they were built-in. Removed from package.json

janedegtiareva:
> I've got a lot of the shell NEP live here:
> https://github.com/nearprotocol/NEPs/blob/6b6e9ce2ab5f9cb36b69accae3970d8ea997a29b/text/0000-near-shell.md#upgradability
> 
> It would be great if this were stored like this, if possible:
> in `~/.near-config/settings.js`:
> 
> ```
> {
>   "metricsOptIn": true
> }
> ```
> 
> There will be other keys in here as well, so we'll want to make sure not to lose any information. (Load file contents as js object, modify that key, write whole thing back to file…)

Yeah that's exactly what's happening

janedegtiareva:
tested: E2E test with making sure event shows up in dashboard for mixpanel. https://mixpanel.com/report/2144016/live#chosen_columns:!('$browser','$city',mp_country_code,distinct_id,'$referring_domain'),column_widths:!(200,100,218,149,217,252,180)

mikedotexe:
Is this outdated now and we're favoring https://github.com/nearprotocol/near-shell/pull/323?

PR #322, opened 2020-04-21
feat: add mainnet config

bowenwang1996:
Add mainnet config for production environment. Also change testnet config urls.

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/322"></a>

### DIFF
--- a/config.js
+++ b/config.js
@@ -6,10 +6,10 @@ function getConfig(env) {
     case 'production':
         return {
             networkId: 'mainnet',
-            nodeUrl: 'https://rpc.nearprotocol.com',
+            nodeUrl: 'https://rpc.mainnet.nearprotocol.com',
             contractName: CONTRACT_NAME,
-            walletUrl: 'https://wallet.nearprotocol.com',
-            helperUrl: 'https://helper.nearprotocol.com',
+            walletUrl: 'https://wallet.mainnet.nearprotocol.com',
+            helperUrl: 'https://helper.mainnet.nearprotocol.com',
         };
     case 'development':
         return {
--- a/config.js
+++ b/config.js
@@ -4,13 +4,20 @@ function getConfig(env) {
     switch (env) {
 
     case 'production':
-    case 'development':
         return {
-            networkId: 'default',
+            networkId: 'mainnet',
             nodeUrl: 'https://rpc.nearprotocol.com',
             contractName: CONTRACT_NAME,
             walletUrl: 'https://wallet.nearprotocol.com',
             helperUrl: 'https://helper.nearprotocol.com',
+        };
+    case 'development':
+        return {
+            networkId: 'testnet',
+            nodeUrl: 'https://rpc.testnet.nearprotocol.com',
+            contractName: CONTRACT_NAME,
+            walletUrl: 'https://wallet.testnet.nearprotocol.com',
+            helperUrl: 'https://helper.testnet.nearprotocol.com',
         };
     case 'devnet':
         return {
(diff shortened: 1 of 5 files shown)


config.js
```diff
@@ -4,14 +4,21 @@ function getConfig(env) {
...
-            networkId: 'default',
+            networkId: 'mainnet',
             nodeUrl: 'https://rpc.nearprotocol.com',
             contractName: CONTRACT_NAME,
             walletUrl: 'https://wallet.nearprotocol.com',
             helperUrl: 'https://helper.nearprotocol.com',
         };
+    case 'development':
+        return {
+            networkId: 'testnet',
```
vgrichina, line 16: maybe makes sense to keep this as `default` to avoid losing keys (though alternatively we can document how to move keys).

config.js
```diff
@@ -4,14 +4,21 @@ function getConfig(env) {
     switch (env) {
 
     case 'production':
-    case 'development':
         return {
-            networkId: 'default',
+            networkId: 'mainnet',
             nodeUrl: 'https://rpc.nearprotocol.com',
             contractName: CONTRACT_NAME,
             walletUrl: 'https://wallet.nearprotocol.com',
```
vgrichina, line 11: this would be a problem given these are URLs for testnet wallet, helper and RPC right now.

I think we might need to do update in steps:
1) update both `production` and `development` to point to testnet URLs. This will need to happen not only in near-shell, but in all examples as well.
2) after 1) we need to disable `wallet.nearprotocol.com`, `helper.nearprotocol.com` and  `rpc.nearprotocol.com` (get everybody who tries hitting them error with instructions to switch to testnet). For Wallet we'll need some key migration strategy.
3) after enough time passed for 2) to affect anybody active enough (and when wallet is ready for mainnet) – we can bring up these endpoints and update `production` to mean `mainnet`.

bowenwang1996: but this means that near-shell will be hard to use for mainnet for a while

bowenwang1996: The plan on the wallet side is to have a landing page for mainnet to inform people that mainnet wallet is not ready yet.

vgrichina: > but this means that near-shell will be hard to use for mainnet for a while

@bowenwang1996 I think we should add explicit `mainnet` and `testnet` as part of (1).

> The plan on the wallet side is to have a landing page for mainnet to inform people that mainnet wallet is not ready yet.

this is not enough if we don't want people to lose their testnet accounts at `wallet.nearprotocol.com` (though if we put on landing page and nobody complains – maybe we shouldn't care :D).

ilblackdragon: I agree, let's start with:
 - rpc.mainnet.nearprotocol.com
 - wallet.mainnet.nearprotocol.com

While we are migrating.

Migrating `default` is indeed important. Let's may be add temporary message that when you call it first time with `default` asks - if you are using mainnet or testnet? If they say `testnet` it renames default into `testnet`.

Ideally we should stop using `NODE_ENV` as it's very weird. And instead adopt something like `near network testnet/mainnet`. Then if network is not selected at the start, it can list all networks and ask to select one.

@bowenwang1996 I think for first 1-2 weeks it's fine if to use MainNet ppl need to set it up. It's going to be mostly us and few initial developers.

bowenwang1996: I have changed the urls to have `mainnet` in them. What should I do about `default`?

config.js
bowenwang1996: I reverted my change

Issue #198, opened 2019-11-20
Key management improvements: current usage of neardev is confusing


janedegtiareva:
Stakewars feedback: 
nearshell always creates a folder of configuration named "neardev" on current directory after running "near login" command , when running nearshell on another folder without "neardev" folder, nearshell's login state would change to unlog, which would cause users to be confused .

Proposed solution is to store keys under .near (with the exception of development projects that keep their keys under neardev).

kcole16:
@janedegtiareva Can this issue be closed?

PR #323, opened 2020-04-21
Metrics events - add utils to support mixpanel integration and persisting user settings in shell.

janedegtiareva:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/323"></a>


utils/eventtracking.js
```diff
@@ -0,0 +1,83 @@
+const MIXPANEL_TOKEN = '9aa8926fbcb03eb5d6ce787b5e8fa6eb';
+var mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
+
+const uuid = require('uuid');
+const chalk = require('chalk');  // colorize output
+const readline = require('readline');
+const settings = require('./settings');
+
+const TRACKING_ENABLED_KEY = 'trackingEnaled';
```
mikedotexe, line 9: typo here :)

mikedotexe:
In standup I heard that we wanted to reach consensus on something? I don't see any details on what is being asked. Looking at the code, this seems great to me. I think adding events to every command is the only way to go. Can't wait to see this live!
Looks like we've got some linting issues, FYI

commands/create-account.js
```diff
@@ -31,6 +32,8 @@ module.exports = {
 };
 
 async function createAccount(options) {
+    await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_START, {});
+    console.log("?")
```
mikedotexe, line 36: ?

utils/settings.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Persistent shell settings
+
+const getShellSettings = () => {
+    const nearPath = path.join(homedir, '.near-config');
+    try {
+        if (!fs.existsSync(nearPath)) {
+            fs.mkdirSync(nearPath);
+        }
+        const shellSettingsPath = path.join(nearPath, 'settings');
```
mikedotexe, line 14: can we have this be `settings.js` please

test/testsettings
```diff
@@ -0,0 +1 @@
+{"trackingEnaled":false}
```
ilblackdragon, line 1: I don't think we should include this file to repo

utils/settings.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs');
+const homedir = require('os').homedir();
+const path = require('path');
+
+
+// Persistent shell settings
+
+const getShellSettings = () => {
+    const nearPath = path.join(homedir, '.near-config');
```
ilblackdragon, line 9: should this be .near or .near-shell or .neardev?

ilblackdragon: also let's make this a constant? same for settings.js

utils/eventtracking.js
```diff
@@ -0,0 +1,83 @@
...
+        const rl = readline.createInterface({
+            input: process.stdin,
+            output: process.stdout
+        });
+        const getEventTrackingConsent = async () => {
+            for (var attempts = 0; attempts < 10; attempts++) {
+                const answer = await new Promise((resolve) => {
+               
+                    rl.question(
+                        chalk`We would like to collect data on near shell usage to improve developer experience. {bold.yellow Would you like to opt in (y/n)?}`,
```
ilblackdragon, line 25: we should probably talk more information about what we are collecting and how we are planning to use this.

utils/settings.js
mikedotexe: > should this be .near or .near-shell or .neardev?

@ilblackdragon I recommended Jane put it here as it fits with the structure I laid out in the NEP.
https://github.com/nearprotocol/NEPs/blob/974887126800be47a58e968fac5b997db7d50f7d/text/0000-near-shell.md#settings-config-and-key-management

```
─ awesome-near-project  ⟵ NEAR dApp
 ├── .near-config       ⟵ Stores shell-experience settings and connection configuration
 │  ├── connections     ⟵ Contains files used to connect and deploy a contract (except keys)
 │  │  ├── default.js   ⟵ Default configuration is for development
 │  │  └── localnet.js  ⟵ Custom connection added by user for localnet
 │  └── settings.js     ⟵ Stores near-shell settings, how shell behaves when run in this project
 └── .near-credentials  ⟵ Previously "neardev" this directory contains private key inforamtion
    └── default         
       └── alice.json   ⟵ NEAR account "alice" has private key information here
```
(the same structure also appears in the home directory)
Based on the [comment](https://github.com/nearprotocol/NEPs/pull/31#issuecomment-606615736) in the shell NEP, I understood that we were not going to put anything in `~/.near` as that's reserved for validator data. We can modify the NEP to be `.neardev` if we want instead of `.near-config`. Currently `neardev` isn't a great name and holds private keys, which the NEP calls to change to a more informative directory `.near-credentials` that folks aren't likely to accidentally revision. Anyway, that's the thought process. Again, we can modify the NEP.

utils/eventtracking.js
janedegtiareva: eagle eye! fixed.

utils/settings.js
janedegtiareva: .json would be more appropriate. I don't want to make it .js because it's not actually a javascript file, but I am on board with .json

test/testsettings
janedegtiareva: How would you propose to handle this in CI if we don't have this file?

utils/eventtracking.js
janedegtiareva: @potatodepaulo could we get some final wording for this?

utils/settings.js
janedegtiareva: moved to consts

utils/eventtracking.js
mikedotexe: How about:
> We would like to collect data on near-shell usage to improve developer experience. We will never send private information. We only collect which commands are run via an anonymous identifier.

test/testsettings
janedegtiareva: removed file and started generating from .sh

utils/eventtracking.js
potatodepaulo: The more text you put in here, the less people are going to read. For now let’s use Mike’s suggestion, and then we can add a link to a data privacy page with a full explanation later.

mikedotexe:
Merge conflict 🎗️

utils/settings.js
ilblackdragon: Do we need configs inside project directory?
I really think we should not store any credentials inside project directory - because that's the easiest way to lose them. E.g. #198 
I also think settings should live in common directory for all projects.

potatodepaulo: @ilblackdragon we don't need credentials, but per-project config is necessary.  At the very least, it's important for making sure that people have updated deps e.g. shell itself. I agree that creds and global config should be in a common dir for everything.

potatodepaulo:
@janedegtiareva looks like it's asking CI to opt in and that's why the build is breaking for this PR.

utils/settings.js
ilblackdragon: @potatodepaulo read the NEP and left comments there - https://github.com/nearprotocol/NEPs/pull/31#pullrequestreview-399439203
High level, I don't think there is need in local configuration folder, especially one that starts with `.` as those directories should be gitignored. `package.json` can contain settings that should be "clonnable".

ilblackdragon:
@janedegtiareva we for sure need to figure out how to not show this in CI - we will have this problem in all examples.

utils/settings.js
janedegtiareva: Can we limit the discussion to user defined settings here in order to get the PR in for RL1?

janedegtiareva:
> @janedegtiareva we for sure need to figure out how to not show this in CI - we will have this problem in all examples.

This was temporary not working due to changing some things due to comments. Back to working again.

janedegtiareva:
> Merge conflict 🎗️

resolved

janedegtiareva:
> @janedegtiareva looks like it's asking CI to opt in and that's why the build is breaking for this PR.

Yeah this was temporarily broken due to commit /addressing comments churn. Back to working now.

utils/eventtracking.js
```diff
@@ -0,0 +1,85 @@
...
+    EVENT_ID_ACCOUNT_STATE_START: 'shell_account_state_start',
+    EVENT_ID_ACCOUNT_STATE_SUCCESS: 'shell_account_state_success',
+    EVENT_ID_DELETE_ACCOUNT_START: 'shell_delete_account_start',
+    EVENT_ID_DELETE_ACCOUNT_SUCCESS: 'shell_delete_account_success',
+    EVENT_ID_ACCOUNT_KEYS_START: 'shell_account_keys_start',
+    EVENT_ID_ACCOUNT_KEYS_SUCCESS: 'shell_account_keys_success',
+    EVENT_ID_TX_STATUS_START: 'shell_tx_status_start',
+    EVENT_ID_TX_STATUS_SUCCESS: 'shell_tx_status_success',
+    EVENT_ID_LOGIN: 'shell_login',
+    EVENT_ID_DEPLOY: 'shell_deploy',
```
mikedotexe, line 81: not used

utils/eventtracking.js
```diff
@@ -0,0 +1,85 @@
...
+    EVENT_ID_DEV_DEPLOY: 'shell_dev_deploy',
```
mikedotexe, line 82: not used

utils/eventtracking.js
```diff
@@ -0,0 +1,85 @@
...
+    // since mixpanel might be used for other components as well.
+    EVENT_ID_ACCOUNT_STATE_START: 'shell_account_state_start',
+    EVENT_ID_ACCOUNT_STATE_SUCCESS: 'shell_account_state_success',
+    EVENT_ID_DELETE_ACCOUNT_START: 'shell_delete_account_start',
+    EVENT_ID_DELETE_ACCOUNT_SUCCESS: 'shell_delete_account_success',
+    EVENT_ID_ACCOUNT_KEYS_START: 'shell_account_keys_start',
+    EVENT_ID_ACCOUNT_KEYS_SUCCESS: 'shell_account_keys_success',
+    EVENT_ID_TX_STATUS_START: 'shell_tx_status_start',
+    EVENT_ID_TX_STATUS_SUCCESS: 'shell_tx_status_success',
+    EVENT_ID_LOGIN: 'shell_login',
```
mikedotexe, line 80: not used

utils/eventtracking.js
janedegtiareva: that's already addressed in the part 2 of this effort which actually does events for all the things. can we let this be to avoid unnecessary churn?

PR #324, opened 2020-04-23
Addresses subdomains in create_account, dev-deploy

mikedotexe:
Addresses https://github.com/nearprotocol/near-shell/issues/307.
Note that there is still an underlying issue when the test tries to deploy a contract:
```
Error:  LackBalanceForState [Error]: The account dev-1587603497974.test.near wouldn't have enough balance to cover storage, required to have 16543700000000000000000
    at Object.parseRpcError (/Users/mike/near/near-shell/node_modules/near-api-js/lib/utils/rpc_errors.js:25:19)
    at Account.signAndSendTransaction (/Users/mike/near/near-shell/node_modules/near-api-js/lib/account.js:114:36)
    at async LocalAccountCreator.createAccount (/Users/mike/near/near-shell/node_modules/near-api-js/lib/account_creator.js:23:9)
    at async createDevAccountIfNeeded (/Users/mike/near/near-shell/commands/dev-deploy.js:67:5)
    at async devDeploy (/Users/mike/near/near-shell/commands/dev-deploy.js:34:23)
    at async Object.handler (/Users/mike/near/near-shell/utils/exit-on-error.js:4:9) {
  type: 'LackBalanceForState',
  index: null,
  amount: '16543700000000000000000',
  signer_id: undefined,
  account_id: 'dev-1587603497974.test.near'
```
Edit: resolved, see Vlad's comment below.

Edit2:

results in:

> 

passes

succeeds

succeeds

results in:

results in:

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/324"></a>

### DIFF
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -1,11 +1,18 @@
-
 const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { KeyPair, utils } = require('near-api-js');
+const NEAR_ENV_SUFFIXES = [
+  'near',
+  'test',
+  'beta',
+  'dev'
+];
+const TLA_MIN_LENGTH = 11;
+const ERROR_INVALID_FORMAT = new Error("Invalid format for account name, please check console for details.")
 
 module.exports = {
     command: 'create_account <accountId>',
-    desc: 'create a new developer account',
+    desc: 'create a new developer account (top-level account 11+ chars, or a subdomain of the masterAccount)',
     builder: (yargs) => yargs
         .option('accountId', {
             desc: 'Unique identifier for the newly created account',
@@ -31,6 +38,26 @@ module.exports = {
 };
 
 async function createAccount(options) {
+    // periods are disallowed in top-level accounts and can only be used for subdomains
+    const splitAccount = options.accountId.split('.');
+    if (splitAccount.length === 2) {
+        // TLA (bob-at-least-maximum-chars.test)
+        if (splitAccount[0].length < TLA_MIN_LENGTH) {
+            console.log(`Top-level accounts (not ending in .near, .test, etc) must be greater than ${TLA_MIN_LENGTH} characters`);
+            throw(ERROR_INVALID_FORMAT);
+        }
+    } else if (splitAccount.length === 3) {
+        // Subdomain (short.alice.near)
+        if (!NEAR_ENV_SUFFIXES.includes(splitAccount[2])) {
+            console.log(`Expected a subdomain account name ending in ${NEAR_ENV_SUFFIXES.join(', ')}. (Example: counter.alice.test)`);
+            throw(ERROR_INVALID_FORMAT);
+        }
+    } else {
+        console.log(`Unexpected account name format. Please use one of these formats:\n` +
+        `1. Top-level name with ${TLA_MIN_LENGTH}+ characters (ex: near-friend.test)\n` +
+        `2. A subdomain account name ending with .${NEAR_ENV_SUFFIXES.join(', .')}. (Example: counter.alice.test)`);
+        throw(ERROR_INVALID_FORMAT);
+    }
     options.initialBalance = utils.format.parseNearAmount(options.initialBalance);
     // NOTE: initialBalance is passed as part of config here
     let near = await connect(options);
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -8,7 +8,6 @@ const NEAR_ENV_SUFFIXES = [
     'dev'
 ];
 const TLA_MIN_LENGTH = 11;
-const ERROR_INVALID_FORMAT = new Error('Invalid format for account name, please check console for details.');
 
 module.exports = {
     command: 'create_account <accountId>',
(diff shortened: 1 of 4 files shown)


vgrichina:
@mikedotexe 
> Note that there is still an underlying issue when the test tries to deploy a contract:

I think it should be fixed by this https://github.com/nearprotocol/near-shell/commit/7e31bfb9a635d34fc61de245401dcc4d4f4d256f, try rebasing/merging with `master`

mikedotexe:
Update: I think I am not understanding that a TLA is according to @bowenwang1996 who had a comment here:
https://github.com/nearprotocol/near-shell/issues/307#issuecomment-618478994

commands/create-account.js
```diff
@@ -32,6 +38,26 @@ module.exports = {
 
 async function createAccount(options) {
```
potatodepaulo, line 39: You should add a check that the suffix matches the network that the user is talking to.

potatodepaulo: other than that, looks good.

commands/create-account.js
```diff
@@ -1,11 +1,17 @@
-
 const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { KeyPair } = require('near-api-js');
+const NEAR_ENV_SUFFIXES = [
+    'near',
```
vgrichina, line 5: I think this should go to config. Maybe even just reuse `masterAccount` for this, but use `helperUrl` for account creation when available.

commands/create-account.js
```diff
@@ -32,6 +38,26 @@ module.exports = {
...
     // NOTE: initialBalance is passed as part of config here, parsed in middleware/initial-balance
+    // periods are disallowed in top-level accounts and can only be used for subdomains
+    const splitAccount = options.accountId.split('.');
+    if (splitAccount.length === 2) {
+        // TLA (bob-at-least-maximum-chars.test)
+        if (splitAccount[0].length < TLA_MIN_LENGTH) {
+            console.log(`Top-level accounts (not ending in .near, .test, etc) must be greater than ${TLA_MIN_LENGTH} characters`);
+            return;
+        }
+    } else if (splitAccount.length === 3) {
```
vgrichina, line 49: length isn't limited to 3, can be easily `foo.bar.baz.alice.near` if you have `bar.baz.alice.near` as `masterAccount`

commands/dev-deploy.js
```diff
@@ -67,7 +67,7 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
         }
     }
 
-    const accountId = `dev-${Date.now()}`;
+    const accountId = `dev-${Date.now()}.test.near`;
```
vgrichina, line 70: I think we should take `masterAccount` from config, suffix here not necessarily `test.near`.

commands/create-account.js
mikedotexe: `helperUrl`? I thought that is only used in the dev-deploy. As far as I know, the `create_account` command will always need a `masterAccount` right?

vgrichina: it is generally used when you create accounts using `near.createAccount` https://github.com/near/near-api-js/blob/f1dcf73ff3549fe7bc7caf35c7c6e43fbc4a2703/src/near.ts#L59
https://github.com/near/near-api-js/blob/f1dcf73ff3549fe7bc7caf35c7c6e43fbc4a2703/src/near.ts#L37

Though really the only hard statement here is that name suffix should be configurable, I'm not sure it should be universally be the same as `masterAccount`.

mikedotexe: Gotcha, interesting. Currently if you try passing just the  `helperUrl` yargs is set to fail because `masterAccount` is required.

`./bin/near create_account mike.asdf.test --accountHelper https://helper.testnet.nearprotocol.com`
gives
>Missing required argument: masterAccount
I feel like that'd be a good enhancement for another issue. Is it cool if we do this in another PR?

mikedotexe: I do like the idea of using config for those, by the way, and I'll be pushing up a change that checks config's `networkId` against it.

vgrichina: >  and I'll be pushing up a change that checks config's networkId against it.

not sure what do you mean

commands/create-account.js
mikedotexe: thanks, this is helpful feedback

commands/create-account.js
vgrichina: > I feel like that'd be a good enhancement for another issue. Is it cool if we do this in another PR?

sure, my main point here is just so that we avoid hardcoding suffixes in shell. If they are in config – they can easily be altered for a given project, etc. Probably should also be passable as argument (again so that it's easy to override when needed).

mikedotexe: I agree, I just pushed up a commit that essentially _suggests_ to the user that most of the time if they're on betanet, the account will end in .beta. But it's not blocking. We can't make it a hard requirement because of the custom TLAs, too. Good point

mikedotexe:
@vgrichina when you get a sec, this is ready for re-review. I just moved it out of Draft, so want to ping as I don't know if Github alerts reviewers upon that action.

PR #331, opened 2020-04-28
Hotfix/307 subdomains

mikedotexe:
Second attempt at issue:
https://github.com/nearprotocol/near-shell/issues/307

First attempt was here:
https://github.com/nearprotocol/near-shell/pull/324

It was my mistake in not testing it properly and we reverted that PR.

The new item here is during dev-deploy, we check to see what environment the user is on (betanet, testnet, etc) and if it's using a `helperUrl` determine what the name of the account is that'll be funding the creation. So we split the `helperUrl`. Basically a newly created account can either be from a `masterAccount` or from a `helperUrl` and we need to determine what the account suffix will be.

With the latest commit this kind of thing will work, when before it would fail:
`NODE_ENV=ci yarn near dev-deploy`
^ Note: I ran this in the `guest-book` example after I copy/pasted shell's `config.js` to guest-book's `src/config.js` so it knows about betanet and the new URLs.

So above is the new stuff.
I'll paste the description from the other PR, which is the old stuff:

results in:

> 

passes

succeeds

succeeds

results in:

results in:

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/331"></a>

### DIFF
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -48,7 +48,7 @@ async function devDeploy(options) {
     await eventtracking.track(eventtracking.EVENT_ID_DEV_DEPLOY_END, { node: options.nodeUrl, success: true });
 }
 
-async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount }) {
+async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount, helperUrl }) {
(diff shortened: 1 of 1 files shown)


mikedotexe:
closed, please see https://github.com/nearprotocol/near-shell/pull/332

Issue #307, opened 2020-04-06
Indicate to users created accounts will be subdomains


kcole16:
### Overview
Once https://github.com/nearprotocol/nearcore/issues/2292 lands, near-shell should indicate to users that `near create_account` can only create:
- Subdomains of the sending account `--masterAccount` 
- TLDs with >= 11 characters.

**Example:**
Using `kendall.near`, I can create accounts like `test.kendall.near` and `hasatleasteleven`.

cc @potatodepaulo @evgenykuzyakov

bowenwang1996:
It seems that top level account names that are longer than 11 characters will still be allowed.

ilblackdragon:
We bumping account names that are allowed to 32 characters FYI.

mikedotexe:
I'm getting `CreateAccountNotAllowed` when I am trying this. Is this just disabled on testnet and fixed in the linked nearcore issue?

bowenwang1996:
@mikedotexe well, `hihihihihihi.test` is unfortunately not a TLA.

mikedotexe:
Darn, we need to reopen this and fix some issues Vlad saw this dev-deploy. My mistake for not catching it.

PR #332, opened 2020-04-28
use subaccounts and provide suggestions, fix tests

mikedotexe:
(Needed to make a fresh branch)
Second attempt at issue:
https://github.com/nearprotocol/near-shell/issues/307

First attempt was here:
https://github.com/nearprotocol/near-shell/pull/324

It was my mistake in not testing it properly and we reverted that PR.

The new item here is during dev-deploy, we check to see what environment the user is on (betanet, testnet, etc) and if it's using a `helperUrl` determine what the name of the account is that'll be funding the creation. So we split the `helperUrl`. Basically a newly created account can either be from a `masterAccount` or from a `helperUrl` and we need to determine what the account suffix will be.

With the latest commit this kind of thing will work, when before it would fail:
`NODE_ENV=ci yarn near dev-deploy`
^ Note: I ran this in the `guest-book` example after I copy/pasted shell's `config.js` to guest-book's `src/config.js` so it knows about betanet and the new URLs.

So above is the new stuff.
I'll paste the description from the other PR, which is the old stuff:

results in:

> 

passes

succeeds

succeeds

results in:

results in:

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/331"></a>

### DIFF
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -3,10 +3,18 @@ const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { KeyPair } = require('near-api-js');
 const eventtracking = require('../utils/eventtracking');
+const NEAR_ENV_SUFFIXES = {
+    production: 'near',
+    default: 'testnet',
+    development: 'testnet',
+    devnet: 'devnet',
+    betanet: 'betanet'
+};
+const TLA_MIN_LENGTH = 32;
 
 module.exports = {
     command: 'create_account <accountId>',
-    desc: 'create a new developer account',
+    desc: 'create a new developer account (subaccount of the masterAccount, ex: app.alice.test)',
     builder: (yargs) => yargs
         .option('accountId', {
             desc: 'Unique identifier for the newly created account',
@@ -34,6 +42,39 @@ module.exports = {
 async function createAccount(options) {
     await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_START, { nodeUrl: options.nodeUrl });
     // NOTE: initialBalance is passed as part of config here, parsed in middleware/initial-balance
+    // periods are disallowed in top-level accounts and can only be used for subaccounts
+    const splitAccount = options.accountId.split('.');
+
+    const splitMaster = options.masterAccount.split('.');
+    const masterRootTLA = splitMaster[splitMaster.length - 1];
+    if (splitAccount.length === 1) {
+        // TLA (bob-with-at-least-maximum-characters)
+        if (splitAccount[0].length < TLA_MIN_LENGTH) {
+            console.log(`Top-level accounts must be greater than ${TLA_MIN_LENGTH} characters.\n` +
+              'Note: this is for advanced usage only. Typical account names are of the form:\n' +
+              'app.alice.test, where the masterAccount shares the top-level account (.test).'
+            );
+            return;
+        }
+    } else if (splitAccount.length > 1) {
+        // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
+        // Check that master account TLA matches
+        const accountRootTLA = splitAccount[splitAccount.length - 1];
+        const accountTLA = splitAccount.filter((n, i) => i !== 0).join('.');
+        if (accountTLA !== options.masterAccount) {
+            console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
+            return;
+        }
+
+        // Rules apply for environments except local, test, ci, ci-staging
+        if (Object.keys(NEAR_ENV_SUFFIXES).includes(options.networkId)) {
+            // Recommend that the TLA matches the expected network in most cases
+            const networkTLA = NEAR_ENV_SUFFIXES[options.networkId];
+            if (networkTLA !== masterRootTLA || networkTLA !== accountRootTLA) {
+                console.log(`NOTE: In most cases, when connected to "${options.networkId}" account and masterAccount will end in ".${networkTLA}"`);
+            }
+        }
+    }
     let near = await connect(options);
     let keyPair;
     let publicKey;
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -49,7 +49,7 @@ async function createAccount(options) {
     if (splitAccount.length === 1) {
         // TLA (bob-with-at-least-maximum-characters)
         if (splitAccount[0].length < TLA_MIN_LENGTH) {
-            console.log(`Top-level accounts must be greater than ${TLA_MIN_LENGTH} characters.\n` +
+            console.log(`Top-level accounts must be at least ${TLA_MIN_LENGTH} characters.\n` +
               'Note: this is for advanced usage only. Typical account names are of the form:\n' +
               'app.alice.test, where the masterAccount shares the top-level account (.test).'
             );
@@ -58,15 +58,16 @@ async function createAccount(options) {
     } else if (splitAccount.length > 1) {
         // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
         // Check that master account TLA matches
-        const accountTLA = splitAccount.filter((n, i) => i !== 0).join('.');
-        if (accountTLA !== options.masterAccount) {
+        if (!options.accountId.endsWith(`.${options.masterAccount}`)) {
             console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
             return;
         }
 
         // Warn user if account seems to be using wrong network, where TLA is captured in config
-        if (Object.prototype.hasOwnProperty.call(options, 'tla') && masterRootTLA !== options.tla) {
-            console.log(`NOTE: In most cases, when connected to network "${options.networkId}", masterAccount will end in ".${options.tla}"`);
+        // TODO: when "network" key is available, revisit logic to determine if user is on proper network
+        // See: https://github.com/near/near-shell/issues/387
+        if (options.helperAccount && masterRootTLA !== options.helperAccount) {
+            console.log(`NOTE: In most cases, when connected to network "${options.networkId}", masterAccount will end in ".${options.helperAccount}"`);
         }
     }
     let near = await connect(options);
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -3,13 +3,7 @@ const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { KeyPair } = require('near-api-js');
 const eventtracking = require('../utils/eventtracking');
-const NEAR_ENV_SUFFIXES = {
-    production: 'near',
-    default: 'testnet',
-    development: 'testnet',
-    devnet: 'devnet',
-    betanet: 'betanet'
-};
+// Top-level account (TLA) is testnet for foo.alice.testnet
 const TLA_MIN_LENGTH = 32;
 
 module.exports = {
@@ -59,20 +53,15 @@ async function createAccount(options) {
     } else if (splitAccount.length > 1) {
         // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
         // Check that master account TLA matches
-        const accountRootTLA = splitAccount[splitAccount.length - 1];
         const accountTLA = splitAccount.filter((n, i) => i !== 0).join('.');
         if (accountTLA !== options.masterAccount) {
             console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
             return;
         }
 
-        // Rules apply for environments except local, test, ci, ci-staging
-        if (Object.keys(NEAR_ENV_SUFFIXES).includes(options.networkId)) {
-            // Recommend that the TLA matches the expected network in most cases
-            const networkTLA = NEAR_ENV_SUFFIXES[options.networkId];
-            if (networkTLA !== masterRootTLA || networkTLA !== accountRootTLA) {
-                console.log(`NOTE: In most cases, when connected to "${options.networkId}" account and masterAccount will end in ".${networkTLA}"`);
-            }
+        // Warn user if account seems to be using wrong network, where TLA is captured in config
+        if (Object.prototype.hasOwnProperty.call(options, 'tla') && masterRootTLA !== options.tla) {
+            console.log(`NOTE: In most cases, when connected to network "${options.networkId}", masterAccount will end in ".${options.tla}"`);
         }
     }
     let near = await connect(options);
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -50,7 +50,7 @@ async function devDeploy(options) {
     await eventtracking.track(eventtracking.EVENT_ID_DEV_DEPLOY_END, { node: options.nodeUrl, success: true });
 }
 
-async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount, helperUrl }) {
+async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount, helperUrl, tla }) {
     // TODO: once examples and create-near-app use the dev-account.env file, we can remove the creation of dev-account
     // https://github.com/nearprotocol/near-shell/issues/287
     const accountFilePath = `${PROJECT_KEY_DIR}/dev-account`;
@@ -75,12 +75,12 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init, maste
         }
     }
     let accountId;
-    if (typeof masterAccount === 'undefined' && helperUrl) {
-        // determine "faucet" account name of environment from helperUrl
-        const splitHelper = helperUrl.split('.');
-        accountId = `dev-${Date.now()}.${splitHelper[1]}`;
-    } else {
+    if (typeof masterAccount === 'undefined' && helperUrl && tla) {
+        accountId = `dev-${Date.now()}.${tla}`;
+    } else if (masterAccount) {
         accountId = `dev-${Date.now()}.${masterAccount}`;
+    } else {
+        console.error('Expecting helperUrl and/or tla flag, please provide those.');
     }
 
     const keyPair = await KeyPair.fromRandom('ed25519');
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -51,7 +51,7 @@ async function devDeploy(options) {
     await eventtracking.track(eventtracking.EVENT_ID_DEV_DEPLOY_END, { node: options.nodeUrl, success: true });
 }
 
-async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount, helperUrl, helperAccount }) {
+async function createDevAccountIfNeeded({ near, keyStore, networkId, init, masterAccount }) {
     // TODO: once examples and create-near-app use the dev-account.env file, we can remove the creation of dev-account
     // https://github.com/nearprotocol/near-shell/issues/287
     const accountFilePath = `${PROJECT_KEY_DIR}/dev-account`;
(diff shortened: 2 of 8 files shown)


mikedotexe:
I messaged with Vlad about adding the `masterAccount` info to the config. That's not going to work as far as I can see. We need **either** the `masterAccount` or the `helperUrl` in order to create a new account. if you provide the `masterAccount` it will look locally for the key.
I'm afraid I need to pass off this task to someone else, or else be minimally involved.

commands/create-account.js
```diff
@@ -34,6 +42,39 @@ module.exports = {
 async function createAccount(options) {
     await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_START, { nodeUrl: options.nodeUrl });
     // NOTE: initialBalance is passed as part of config here, parsed in middleware/initial-balance
+    // periods are disallowed in top-level accounts and can only be used for subaccounts
+    const splitAccount = options.accountId.split('.');
+
+    const splitMaster = options.masterAccount.split('.');
+    const masterRootTLA = splitMaster[splitMaster.length - 1];
+    if (splitAccount.length === 1) {
+        // TLA (bob-with-at-least-maximum-characters)
```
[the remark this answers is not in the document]
janedegtiareva: oh I see, we are actually using it with < after. So app.alice.test is actually invalid according to this. Let's update the comment so that it reflects the real situation?

commands/dev-deploy.js
```diff
@@ -69,8 +69,15 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
             }
         }
     }
+    let accountId;
+    if (typeof masterAccount === 'undefined' && helperUrl) {
```
janedegtiareva, line 73: Is the code that provides a helpful message for prod in this PR, or was this some other PR?

mikedotexe: This isn't that code, but I won't have time for that in this PR turns out. Let's just consider that another small issue in my opinion, and consider this PR complete.

commands/create-account.js
```diff
@@ -3,10 +3,18 @@ const exitOnError = require('../utils/exit-on-error');
...
 const { KeyPair } = require('near-api-js');
 const eventtracking = require('../utils/eventtracking');
+const NEAR_ENV_SUFFIXES = {
+    production: 'near',
+    default: 'testnet',
+    development: 'testnet',
+    devnet: 'devnet',
+    betanet: 'betanet'
+};
+const TLA_MIN_LENGTH = 32;
```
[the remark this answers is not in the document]
mikedotexe: top-level-account as mentioned in #307 which this PR is linked to as the issue and has details

commands/create-account.js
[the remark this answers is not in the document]
mikedotexe: no, I believe this is correct. @potatodepaulo can you chime in here

commands/create-account.js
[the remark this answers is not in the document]
mikedotexe: I have this in the Pull Request description:
> Top-level accounts must be greater than 32 characters

janedegtiareva:
I am ok working on the code docs in a follow up PR. Basically todos: 
* the comments for this PR are awesome, but it would be also great to include them in the actual code for the sake of the developer who works in that code a year from now and does not see the PR
* I think it would be good to make an automated test out of the test cases that you listed. It would be a great code coverage improvement!

commands/create-account.js
```diff
@@ -3,10 +3,18 @@ const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { KeyPair } = require('near-api-js');
 const eventtracking = require('../utils/eventtracking');
+const NEAR_ENV_SUFFIXES = {
```
vgrichina, line 6: This for sure shouldn't be hardcoded in shell. IMO makes sense to pass this as `--masterAccount` from config/options, as they have to match basically.

potatodepaulo: Bump on this. We could also get from config, then finally set a default since network is set there.

mikedotexe: Let's chat about this. I'm not sure what is being suggested here. Adding a key for `TLA` to the `config.js`?
As a reminder, since I've taken a while to get back to this, what we're aiming to accomplish is this:
**A user tries to create an account with a betanet account while connected to testnet.**
(This PR has a failure for when the TLA of the new account and master account are off, so that's not the issue at hand.)
If we're going to look at the network the user is connected to and say, "I see you're on testnet (aka `networkId` of **default**) but you're trying to create an account with a .betanet username, watch out"
also, "I see you're on production (aka `networkId` of **mainnet**) but are not using a .near username, watch out"
And since mainnet can have custom TLAs this is going to be just a warning, I believe.
Since the `config.js`'s `networkId` doesn't correspond to the TLA, what's the change here?
Do we change `config.js`'s `default` to be `testnet`?

mikedotexe: okay @vgrichina and @potatodepaulo I think I pushed up a change that will make sense here. Added `tla` key to the `config.js` file. Since we're in `strict` mode, I had to add another (hidden) flag into yargs to accept it. Then I check to see whether the network has a specified TLA, and if it does, check against the `masterAccount` when doing `create_account`

commands/create-account.js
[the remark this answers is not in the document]
mikedotexe: added a comment about tla with an example

commands/create-account.js
```diff
@@ -34,6 +36,34 @@ module.exports = {
...
+        // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
+        // Check that master account TLA matches
+        const accountTLA = splitAccount.filter((n, i) => i !== 0).join('.');
+        if (accountTLA !== options.masterAccount) {
+            console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
+            return;
+        }
+
+        // Warn user if account seems to be using wrong network, where TLA is captured in config
+        if (Object.prototype.hasOwnProperty.call(options, 'tla') && masterRootTLA !== options.tla) {
```
vgrichina, line 63: @mikedotexe is there any reason why we need to treat `tla` differently than `masterAccount`? Like it's not that all accounts on mainnet will have to end with `.near`.

It's just when you want to create `account_name.near` you need to have access to `near` as master account (either directly or through contract helper).

commands/dev-deploy.js
```diff
@@ -74,8 +74,15 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
...
+        // determine "faucet" account name of environment from helperUrl
+        const splitHelper = helperUrl.split('.');
+        accountId = `dev-${Date.now()}.${splitHelper[1]}`;
```
vgrichina, line 81: this seems to be inconsistent with `tla` logic added in other part.

I feel it will all be much easier if we just always use `dev-${Date.now()}.${masterAccount}`;

vgrichina:
I'll do today's cut without this PR for now as I think we need to update project templates as well for anything that requires `src/config.js` changes.

mikedotexe:
> I'll do today's cut without this PR for now as I think we need to update project templates as well for anything that requires `src/config.js` changes.

Sounds good, we can test this some more. I did `yarn link` in this project, went into my Guest Book example, did `yarn link near-shell` (confirmed symlink in `node_modules/.bin/near`) and was able to run `yarn near create_account temp2.mike.testnet --masterAccount mike.testnet`

Looks like `dev-deploy` is having an issue, I'll have to look at this later:
>BadRequestError: Error: The new account_id dev-1590114660184.onrender can't be created by testnet

vgrichina:
> Looks like dev-deploy is having an issue, I'll have to look at this later:

I think it's related to https://github.com/near/near-shell/pull/332#issuecomment-632444155

commands/create-account.js
vgrichina: > Since we're in strict mode, I had to add another (hidden) flag into yargs to accept it. 

I see no need to hide argument for this, if it's in config – it makes sense to document it as argument. It's pretty convenient to have everything set in config overridable from command line.

Also as commented here (https://github.com/near/near-shell/pull/332#discussion_r428976098) I think we can use the same `--masterAccount` key for this, as I don't see why you'd want to set these differently.

If we want to use it to make difference between master account used directly and the one used through helper – maybe it should be smth like `--helperAccount` to make clear it is related? As it for example can be e.g. `stardust.near` if Stardust is hosting their own helper to create accounts for users, but doesn't want to purchase TLD.

mikedotexe:
> > Looks like dev-deploy is having an issue, I'll have to look at this later:
> 
> I think it's related to [#332 (comment)](https://github.com/near/near-shell/pull/332#issuecomment-632444155)

This is a quote and a link to the same comment, I'll probably be able to figure out what you meant though.

commands/create-account.js
mikedotexe: So the purpose of this check is to send a console log (not error) when people try to use `masterAccount` from a different network.
In an ideal world (and I included this in the shell NEP) we would have `{"account_id":…, "networkId":…}` so we can check if they're trying to use the wrong network.
There have been a lot of people (especially validators/stakers) that have forgotten to use the `NODE_ENV=betanet` (or `NEAR_ENV=betanet`) so the only purpose of this is to make sure they're on the right network.
I believe what you're talking about is a check that is right above this piece of code. Right before this we check if the new account has the same TLA as the masterAccount like you're suggesting. That check will `return` unlike this one which is just a "NOTE"

commands/dev-deploy.js
mikedotexe: So `dev-deploy` can use *either* `--helperUrl` which is what it'll be 99% of the time, **or** you can use `--masterAccount`
Am I misunderstanding `dev-deploy` or are you a bit confused. I know this PR has been open for a while and my brain is straining at this point, too.

mikedotexe: I am, however, adding `tla` instead of parsing the `helperUrl`. I definitely missed updating this file after adding that new key

commands/create-account.js
mikedotexe: I like the idea of removing the `hide` from `tla`
While I'm at it, I'm removing all the hidden flags, pushing up a commit soon.

vgrichina:
> > > Looks like dev-deploy is having an issue, I'll have to look at this later:
> > 
> > 
> > I think it's related to [#332 (comment)](https://github.com/near/near-shell/pull/332#issuecomment-632444155)
> 
> This is a quote and a link to the same comment, I'll probably be able to figure out what you meant though.

oops, I wanted to link https://github.com/near/near-shell/pull/332#discussion_r428976690

vgrichina:
> I messaged with Vlad about adding the masterAccount info to the config. That's not going to work as far as I can see. We need either the masterAccount or the helperUrl in order to create a new account. if you provide the masterAccount it will look locally for the key.

It's not like this logic is set in stone. We can and should update near-api-js if necessary. It's pretty convenient to do changes in both places at the same time if you use `yarn link` (https://classic.yarnpkg.com/en/docs/cli/link/#toc-yarn-link-in-package-you-want-to-link)

mikedotexe:
@vgrichina I'm unsure what approach you envision here. In other news, I recently used `yarn link` which is great. What I believe your suggestion is:
just use `dev-${Date.now()}.${masterAccount};` when doing `dev-deploy`
my reaction to that:
if someone uses `helperUrl` (which is most of the time) then there's no `masterAccount` and a dev account will be created with the TLA account, essentially.
It seems like we've gone back and forth on this a while, but I feel like I'm missing some basic about what you'd like to see. Can you please spell it out for me?

commands/create-account.js
vgrichina: > In an ideal world (and I included this in the shell NEP) we would have {"account_id":…, "networkId":…} so we can check if they're trying to use the wrong network.

let's create issue for this and link in `TODO` from here.

commands/create-account.js
```diff
@@ -39,6 +41,34 @@ module.exports = {
...
+            console.log(`Top-level accounts must be greater than ${TLA_MIN_LENGTH} characters.\n` +
+              'Note: this is for advanced usage only. Typical account names are of the form:\n' +
+              'app.alice.test, where the masterAccount shares the top-level account (.test).'
+            );
+            return;
+        }
+    } else if (splitAccount.length > 1) {
+        // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
+        // Check that master account TLA matches
+        const accountTLA = splitAccount.filter((n, i) => i !== 0).join('.');
```
vgrichina, line 61: nit: `splitAccount.slice(1).join('.')`

or just check `if (!accountId.endsWith(`.${masterAccount}`)

commands/create-account.js
```diff
@@ -39,6 +41,34 @@ module.exports = {
...
+        if (accountTLA !== options.masterAccount) {
+            console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
+            return;
+        }
+
+        // Warn user if account seems to be using wrong network, where TLA is captured in config
```
vgrichina, line 67: nit: `Object.prototype.hasOwnProperty.call(options, 'tla')` -> `options.tla` ? Do not think want to treat `tla: null` or `tla: ''` as `tla` being present.

vgrichina:
@mikedotexe looks good overall, but I have few concerns:
- I agree with @janedegtiareva that `TLA_MIN_LENGTH = 32` doesn't look right. Is this coming from some nearcore requirement? Where is the source for this? 
- `tla` is pretty obscure acronym. Should we change to `helperAccount` so that it is at least clear this is account name and somehow related to helper web service? 
- feel like most of this logic should actually live in `near-api-js`. Whoever is using `near.createAccount` would run into same issues as in near-shell. We can ship it in near-shell first if easier, but ultimately it makes sense to move.

mikedotexe:
> @mikedotexe looks good overall, but I have few concerns:
> 
> * I agree with @janedegtiareva that `TLA_MIN_LENGTH = 32` doesn't look right. Is this coming from some nearcore requirement? Where is the source for this?

Illia mentioned the 32 character limit here:
https://github.com/near/near-shell/issues/307#issuecomment-612922970
This is from the issue that's linked in this PR's description. @potatodepaulo has more context on this, and he filled me in.

> * `tla` is pretty obscure acronym. Should we change to `helperAccount` so that it is at least clear this is account name and somehow related to helper web service?

Okay, I can push that up

> * feel like most of this logic should actually live in `near-api-js`. Whoever is using `near.createAccount` would run into same issues as in near-shell. We can ship it in near-shell first if easier, but ultimately it makes sense to move.

It's broken out in `near-api-js` here into two functions, which I don't find confusing, but we can tweak them, sure:
https://github.com/near/near-api-js/blob/master/src/account_creator.ts#L14-L55

vgrichina:
> It's broken out in near-api-js here into two functions, which I don't find confusing, but we can tweak them, sure:

I'm referring to this entry point:
https://github.com/near/near-api-js/blob/25e7e60203965aee062ca19f352ff765859896ef/src/near.ts#L62

I feel like if you use it programmatically in a wrong way – it would make sense if it results same/similar errors as in this PR.

vgrichina:
> @vgrichina I'm unsure what approach you envision here. In other news, I recently used `yarn link` which is great. What I believe your suggestion is:
> just use `dev-${Date.now()}.${masterAccount};` when doing `dev-deploy`
> my reaction to that:
> if someone uses `helperUrl` (which is most of the time) then there's no `masterAccount` (according to `near-shell`, of course there is behind the scenes) and a dev account will be created with the TLA account, essentially.

It only made sense if combined with other suggestion (use `--masterAccount` in place of `--tla`). However having separate `--masterAccount` and `--helperAccount` also makes sense to me. Main problem was leftover `helperUrl` parsing I assume.

mikedotexe:
okay, I think I've addressed all these concerns @vgrichina

vgrichina:
@mikedotexe I've tested this on GitPod as following:

1) `yarn link`
2) `npx create-near-app test-near-app`
3) `cd test-near-app`
4) `yarn link near-shell`
5) `yarn dev` – this fails:

```
 Expecting helperUrl and/or helperAccount flag, please provide those.
Error:  BadRequestError: TypeError: Cannot read property 'constructor' of undefined
    at Object.fetchJson (/workspace/near-shell/node_modules/near-api-js/lib/utils/web.js:47:36)
    at async UrlAccountCreator.createAccount (/workspace/near-shell/node_modules/near-api-js/lib/account_creator.js:42:9)
    at async createDevAccountIfNeeded (/workspace/near-shell/commands/dev-deploy.js:88:5)
    at async devDeploy (/workspace/near-shell/commands/dev-deploy.js:44:23)
    at async Object.handler (/workspace/near-shell/utils/exit-on-error.js:4:9)
```

We probably shouldn't be requiring `helperAccount` until examples are updated. Or maybe you have some other rollout strategy in mind?

mikedotexe:
> We probably shouldn't be requiring `helperAccount` until examples are updated. Or maybe you have some other rollout strategy in mind?

Thanks, Vlad. My `yarn link` was done before this change, I should have checked again. I'm thinking it makes sense to add a friendly warning in the console and proceed to parsing the `helperUrl` as we did before. How does that sound?

vgrichina:
>  I'm thinking it makes sense to add a friendly warning in the console and proceed to parsing the `helperUrl` as we did before. How does that sound?

I think it makes sense to output warning with instructions to update using `src/config.js` from here. Then just proceed to create dev- account without suffix (in this release).

After this we can safely make near-shell release and update examples/templates. Once examples are updated – makes sense to cut another release making this an error and disable creating accounts without suffix in nearcore.

vgrichina:
@mikedotexe just realized that given we can have top level accounts with 32+ characters – another option to avoid breaking examples/templates is just to make sure `dev-deploy` creates long enough names. E.g. something like this `dev-deploy-1590636968030-1000000` (where makes sense to generate last part from `Math.random()` to reduce probability of collision.

Also I realized there is another important thing we'll need to update - the way names are generated in Jest test environment (if it is still used in examples at least):

https://github.com/near/near-shell/blob/a645203bb6427c66c3a8a06e1c42327c4f80eedc/test_environment.js#L21

mikedotexe:
Great idea @vgrichina. I used a random number with at least 7 digits for when they haven't given the flag `--masterAccount` and it works.
I also did the same thing in the test_environment file.
Tested this by `yarn link near-shell`, went into guest book (which has some tests like you mentioned` and ran `yarn test` successfully, and `yarn near dev-deploy` and `yarn dev` etc. Good thinking.

vgrichina:
@mikedotexe 

`create_account` doesn't seem to work as expected?

```
$ bin/near create_account test-create-shell

Missing required argument: masterAccount
```

```
$ bin/near create_account test-create-shell.testnet

Missing required argument: masterAccount
```

mikedotexe:
> @mikedotexe
> 
> `create_account` doesn't seem to work as expected?
> 
> ```
> $ bin/near create_account test-create-shell
> 
> Missing required argument: masterAccount
> ```
> 
> ```
> $ bin/near create_account test-create-shell.testnet
> 
> Missing required argument: masterAccount
> ```

This is how it's always been, are you expecting it to automatically use the `helperUrl`?

vgrichina:
> This is how it's always been,

Yeah, looks so. I think it's a bug though as `near.createAccount` would work fine with same config.

How can I test this codepath otherwise?
https://github.com/near/near-shell/blob/ee4c5413940e6f394e5ce1ed606927f85b50978e/commands/create-account.js#L99-L104

> are you expecting it to automatically use the `helperUrl`?

yes

> Are you suggesting we allow people to create testnet accounts via `near-shell`? What about all the work being done in Wallet to abate people spamming to system to collect tokens?

We allow users to create accounts using helper through `near dev-deploy` and `test_environment.js`  (or `near.createAccount`) anyway, so this doesn't really change much. Faucet will have to do some kind of rate limiting in any case.

vgrichina:
> How can I test this codepath otherwise?

ah wait, just realized it actually requires user to specify `--masterAccount` anyway

mikedotexe:
> @mikedotexe so this is approved (so you can merge), how do you want to proceed?
> 
> 1. merge this already then we can allow creating accounts through helper in follow up (I can do it if you haven't started)
> 2. wait for it and merge as part of this PR (if you already started work)

I'll merge this, thanks for the bump.

PR #339, opened 2020-04-29
Move default location for credentials to .near-credentials

janedegtiareva:
Also fix an issue that login is not honoring the system-wide keystore
and creating its own version.

The new location for keys is, for example on windows with default env: C:\Users\userid\.near-credentials\default 

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-shell/pull/339"></a>

### DIFF
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -51,8 +51,9 @@ async function devDeploy(options) {
 async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
     // TODO: once examples and create-near-app use the dev-account.env file, we can remove the creation of dev-account
     // https://github.com/nearprotocol/near-shell/issues/287
-    const accountFilePath = `${keyStore.keyDir}/dev-account`;
-    const accountFilePathEnv = `${keyStore.keyDir}/dev-account.env`;
+    console.log(near)
+    const accountFilePath = `neardev/dev-account`;
+    const accountFilePathEnv = `neardev/dev-account.env`;
     if (!init) {
         try {
             // throws if either file is missing
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -51,9 +51,8 @@ async function devDeploy(options) {
 async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
     // TODO: once examples and create-near-app use the dev-account.env file, we can remove the creation of dev-account
     // https://github.com/nearprotocol/near-shell/issues/287
-    console.log(near)
-    const accountFilePath = `neardev/dev-account`;
-    const accountFilePathEnv = `neardev/dev-account.env`;
+    const accountFilePath = 'neardev/dev-account';
+    const accountFilePathEnv = 'neardev/dev-account.env';
     if (!init) {
         try {
             // throws if either file is missing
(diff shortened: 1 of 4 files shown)


janedegtiareva:
Tested: 
* near login
* make sure the key is stored in the appropriate location
* near send

mikedotexe:
> This is a great idea! I was just thinking about something like this yesterday. But I thought the implementation might look a little different and I want to hold this up with a conversation, if I can.
> 
> A couple ideas for changing this:
> 
> 1. Why limit it to only storing credentials in the home directory? What if I want to be logged in with certain credentials most of the time, but override for certain projects/folders? We could make this look for `.near-credentials` in the current directory, and if none found, the next directory up, and so forth up to `~` (or maybe `/` but we'd probably run into permissions issues), and finally, at the end of the chain, the old `./neardev` location.
> 2. Do we only want to store credentials in here? Could we make this a `.nearconfig` and make it use a [toml](https://github.com/toml-lang/toml) format (or toml-like, as with `.gitconfig`)?

I think this is good stop-gap before we approve and implement the NEP where the key management does have an order of lookups described:
https://github.com/nearprotocol/NEPs/blob/6b6e9ce2ab5f9cb36b69accae3970d8ea997a29b/text/0000-near-shell.md#key-management

We're at that odd spot where the NEP isn't "approved" but we could really use some of the functionality. I think this is a PR that will grow into the NEP. For instance, the NEP states that keys be in `.near-credentials` although we're keeping the project's `.neardev` for backwards compatibility. I think this smoothly leans into the future of shell, but will be improved as time goes on. And since the default is to save it to the home directory, this is a step in the right direction.

janedegtiareva:
> Wow, I'm surprised how little code this took when you use MergeKeyStore! Awesome 🙌

Yeah, most of the time was "why is this not working" and figuring out the 2nd change... I went in

> This is a great idea! I was just thinking about something like this yesterday. But I thought the implementation might look a little different and I want to hold this up with a conversation, if I can.
> 
> A couple ideas for changing this:
> 
> 1. Why limit it to only storing credentials in the home directory? What if I want to be logged in with certain credentials most of the time, but override for certain projects/folders? We could make this look for `.near-credentials` in the current directory, and if none found, the next directory up, and so forth up to `~` (or maybe `/` but we'd probably run into permissions issues), and finally, at the end of the chain, the old `./neardev` location.
> 2. Do we only want to store credentials in here? Could we make this a `.nearconfig` and make it use a [toml](https://github.com/toml-lang/toml) format (or toml-like, as with `.gitconfig`)?

There are all good options, the question is more of timing (i.e. specifically do we want to do all that before RL1 or not -- I thought we decided in a quick chat on Tuesday that we want to move to home dir right now). Maybe let's have this discussion in https://github.com/nearprotocol/NEPs/pull/31?

vgrichina:
> Do we only want to store credentials in here? Could we make this a .nearconfig and make it use a toml format (or toml-like, as with .gitconfig)?
 
@chadoh I'm not convinced we should use toml. Other than that I agree with @janedegtiareva that this is out of scope for this PR.  Thanks for getting backwards compatibility for `./neardev`, that's good move before event.

mikedotexe:
@janedegtiareva lint issues, but then ship it, Captain

Issue #239, opened 2020-01-13
Stop abusing NODE_ENV


frol:
`NODE_ENV` is a special environment variable used by Node.js, we should use our own name to avoid the collision.

Also, it would be great to have a regular CLI argument for this feature.

vgrichina:
I think we are using it as intended. People use the same shit to manage say different DB environments in Node, don't see how we are different.

I agree that having a separate CLI argument that explicitly selects NEAR environment might be useful though.

janedegtiareva:
+1 to adding CLI argument.

ilblackdragon:
Why not just NEAR_ENV? 
It's totally possible that people want to have few different NEAR envs and for example have production or development for NODE independently.

ilblackdragon:
Can we prioritize this?...
Especially using `production` ... I went to codebase to figure out why mainnet doesn't work :(

chadoh:
For local dev, I haven't yet wanted to override both RPC and Wallet at the same time, and I think it's strange that our currently-suggested `config.js` assumes that I would.

Also, I find it annoying that we export a `getConfig` function when we could just export the constants themselves. This whole `config.js` seems like a way to export some environment variables to me. See suggested simplification here: https://github.com/near-examples/guest-book/pull/78

I've been wondering if we can move this whole config to a file like `.nearconfig`, and possibly put credentials in this file as well, as suggested here: https://github.com/near/near-shell/pull/339#pullrequestreview-402944666

Such a file could live in a user's home directory and then have specific settings overridden per-directory.

Thoughts?

vgrichina:
> For local dev, I haven't yet wanted to override both RPC and Wallet at the same time, and I think it's strange that our currently-suggested config.js assumes that I would.

Not sure what do you mean. If you are overriding RPC you generally need to override the wallet, otherwise you have a wallet which doesn't work with your RPC.

> Also, I find it annoying that we export a getConfig function when we could just export the constants themselves. This whole config.js seems like a way to export some environment variables to me. See suggested simplification here: near-examples/guest-book#78

We have far more pressing devx issues to solve:
   - support of multiple contracts
   - data migration when upgrading contact
   - batch import/export of contract data
   - error handling being a mess
   - no demo taking advantage of multiple shards
   - etc
I suggest we focus on big picture vs trying to do marginal improvements to config file (which are also a breaking change from POV of everyone already working with us). 

[...]

PR #351, opened 2020-05-05
Allow to use NEAR_ENV to select current env

vgrichina:
Ref https://github.com/near/near-shell/issues/239

### DIFF
--- a/get-config.js
+++ b/get-config.js
@@ -1,13 +1,17 @@
 
 module.exports = function getConfig() {
     const configPath = process.cwd() + '/src/config';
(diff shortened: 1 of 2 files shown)


get-config.js
```diff
@@ -1,13 +1,17 @@
 
 module.exports = function getConfig() {
     const configPath = process.cwd() + '/src/config';
+    const nearEnv = process.env.NODE_ENV || process.env.NODE_ENV || 'development'
```
ilblackdragon, line 4: two times process.env.NODE_ENV?

vgrichina: oops, yeah, not tested yet – will do before pushing release

Issue #358, opened 2020-05-08
"near login" auto-opens wrong Wallet link on Linux


mikedotexe:
There have been a number of folks who have tried using the command "near login" and have seen a red banner at the top. It appears two folks have been on Ubuntu and seen this.
I've installed VMs of three version of Ubuntu and can't seem to reproduce it, but I can reproduce this in Gitpod.

Steps to reproduce: open up a Gitpod instance. (from an example from [near.dev](https://near.dev), for instance) and in the Terminal run `near login` **twice**

The second time, you'll note that the URL that automatically opens in the browser tab has a different public key than the link in the command line. I've spent enough time trying to track this down, that I'm considering this an exotic bug and it's deprioritized. It was a cherry-on-top feature to begin with, and it's unfortunately blocking a lot of our users.

Here are three different cases where this has happened in the past week, screenshots from Slack:
<img width="389" alt="Screenshot 2020-05-08 10 37 38" src="https://user-images.githubusercontent.com/1042667/81432792-62bcbb80-9118-11ea-9af5-33a9b5ea7d46.png">
<img width="327" alt="Screenshot 2020-05-08 10 37 15" src="https://user-images.githubusercontent.com/1042667/81432799-65b7ac00-9118-11ea-8e22-2425c5ce4a5e.png">
<img width="323" alt="Screenshot 2020-05-08 10 36 44" src="https://user-images.githubusercontent.com/1042667/81432805-681a0600-9118-11ea-869b-08fc6c8e3058.png">

vgrichina:
@mikedotexe I cannot reproduce this on GitPod. Can you give more details on how to reproduce?

I think we shouldn't be disabling this flow for GitPod.

vgrichina:
Note that examples need to include this in config for smoother flow:
https://github.com/near/near-shell/blob/7e3f8b1dad129439ab7eb9ba811d659b3042ecfa/.gitpod.yml#L7

mikedotexe:
@vgrichina just to make sure, what was your steps to reproduce? As you see I merged #361 so by default, Gitpod's `master` will work as expected, except you have to enter your id in the terminal.
I followed this procedure to reproduce on Brave and Chrome.

1. Visit: https://github.com/near/near-shell
2. open Gitpod in new tab
3. In the Gitpod Terminal go back to before the merge with:
4. `git checkout 779fab0158c20e0932d3527c33fde2a7b5756bca`
5. Then run `./bin/near login`
6. Follow instructions
[...]

PR #359, opened 2020-05-08
near login: do not open webpage automatically on linux

mikedotexe:
Workaround for issue described here: https://github.com/near/near-shell/issues/358

### DIFF
--- a/index.js
+++ b/index.js
@@ -78,6 +78,7 @@ exports.login = async function (options) {
         // attempt to capture accountId automatically via browser callback
         let tempUrl;
         const isWin = process.platform === 'win32';
+        const isLinux = process.platform === 'linux';
 
         // find a callback URL on the local machine
         try {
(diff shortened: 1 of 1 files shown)


mikedotexe:
@vgrichina I can't find it right now, but I remember we both saw a community member using Ubuntu see this errant behavior even with the change using 302 instead of 301.

vgrichina:
@mikedotexe I don't think we should disable it for everyone on Linux if it happens only in some edge case (i.e. we cannot reproduce ourselves on stock Ubuntu, etc).

Maybe that user just had their Chrome still keep the cache from attempt with unfixed version (most likely explanation). If it happens again we should direct user to clear the cache and see if it helps.

mikedotexe:
> @mikedotexe I don't think we should disable it for everyone on Linux if it happens only in some edge case (i.e. we cannot reproduce ourselves on stock Ubuntu, etc).
> 
> Maybe that user just had their Chrome still keep the cache from attempt with unfixed version (most likely explanation). If it happens again we should direct user to clear the cache and see if it helps.

I disagree. We have consistently heard from people (at least 5-6 in one week) that this is happening. People can click on the link in the command line. We've introduced a bug in the feature that automatically opens a tab. Let's remove that buggy feature and unblock our developers/stakers.

vgrichina:
@mikedotexe there is no plausible explanation for this happening besides Chrome still using old cached URL (because of 301 being used earlier). It's not like wrong link can be opened because of some magical reason.

mikedotexe:
Okay, I guess I'll (re)close this according to the last comment and since we haven't heard of this happening for a couple weeks.

PR #361, opened 2020-05-08
address windows, mac, linux behavior w/ callback and auto-open Wallet

mikedotexe:
Addresses: https://github.com/near/near-shell/issues/358
See the reproduction steps there.
Basically we cannot trust any OS other than Mac to have the callback and auto-open links working consistently.
It was a nice cherry-on-top, but we are causing errant behavior for our users and need to stop it. Unless you're using a Mac.

### DIFF
--- a/index.js
+++ b/index.js
@@ -82,7 +82,8 @@ exports.login = async function (options) {
 
         // find a callback URL on the local machine
         try {
-            if (isMac) { // capture callback is currently not working on windows. This is a workaround to not use it
+            // capture callback isn't working reliably outside of Mac. This is a workaround to not use it
+            if (isMac) {
                 tempUrl = await capture.callback(5000);
                 // if we found a suitable URL, attempt to use it
                 newUrl.searchParams.set('success_url', `http://${tempUrl.hostname}:${tempUrl.port}`);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -77,34 +77,26 @@ exports.login = async function (options) {
...
         // attempt to capture accountId automatically via browser callback
         let tempUrl;
-        const isWin = process.platform === 'win32';
+        // See: https://github.com/near/near-shell/issues/358
+        const isMac = process.platform === 'darwin';
 
         // find a callback URL on the local machine
         try {
-            if (!isWin) { // capture callback is currently not working on windows. This is a workaround to not use it
+            if (isMac) { // capture callback is currently not working on windows. This is a workaround to not use it
```
janedegtiareva, line 85: do you mind updating the comment too?

index.js
```diff
@@ -77,34 +77,26 @@ exports.login = async function (options) {
...
+                // redirect automatically, but do not use the browser callback
+                await openUrl(newUrl);
             }
         } catch (error) {
             // console.error("Failed to find suitable port.", error.message)
             // TODO: Is it? Try triggering error
             // silent error is better here
         }
 
-        // if we found a suitable URL, attempt to use it
```
janedegtiareva, line 93: nice refactor to reduce number of lines! 👍 It's way more readable

index.js
mikedotexe: done, thanks

PR #362, opened 2020-05-09
dev-deploy: create neardev directory if it doesn't exist

mikedotexe:
If there's no `neardev` directory, `near dev-deploy` will fail with:

```
Error:  [Error: ENOENT: no such file or directory, open 'neardev/dev-account'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'neardev/dev-account'
}
error Command failed with exit code 1.
```

This is new behavior as of 0.23.0, let's address this quickly.

Tested by going into guest-book example locally:
`../near-shell/bin/near dev-deploy`

`rm -rf neardev`
then running it again
`rm neardev/dev-account*`
then running it again

### DIFF
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -64,7 +64,10 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
         } catch (e) {
             if (e.code === 'ENOENT') {
                 // Create neardev directory, new account will be created below
-                await mkdir('./neardev');
+                try {
+                    await mkdir('./neardev')
+                } catch (_) {} // if already exists, okay
+
             } else {
                 throw e;
             }
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -54,6 +54,7 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
     // https://github.com/nearprotocol/near-shell/issues/287
     const accountFilePath = 'neardev/dev-account';
     const accountFilePathEnv = 'neardev/dev-account.env';
+    const projectCredentialsPath = './neardev';
     if (!init) {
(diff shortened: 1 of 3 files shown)


mikedotexe:
> I would prefer we did the equivalent of `mkdir -p` in the normal flow, rather than creating the directory in the `catch` block

That won't work for Windows, but I did consider adding the [mkdir package](https://www.npmjs.com/package/mkdirp) for this. I know there's always pushback against adding new dependencies for one small change, so I decided to do it this way.

chadoh:
Right, I said "something like `mkdir -p` because I thought windows might give problems with actual `mkdir -p`. 

I think we could use something like this:

```js
if (!fs.existsSync('./neardev')) await fs.promises.mkdir('./neardev')
```

(Not sure I got that syntax right)

Would this work on Windows?

commands/dev-deploy.js
```diff
@@ -63,7 +64,8 @@ async function createDevAccountIfNeeded({ near, keyStore, networkId, init }) {
             }
         } catch (e) {
             if (e.code === 'ENOENT') {
-                // Ignore as it means new account needs to be created, which happens below
+                // Create neardev directory, new account will be created below
+                if (!existsSync('./neardev')) await mkdir('./neardev');
```
janedegtiareva, line 68: let's pull the dir name into const, so that we don't have string literal duplication

mikedotexe: pushed

vgrichina:
This change https://github.com/near/near-shell/pull/362/commits/d4262019a84a471f5343d55e127cf331d58c0c91 only works well if all the occurrences are replaced:
https://github.com/near/near-shell/search?l=JavaScript&q=neardev

PR #363, opened 2020-05-11
Use temporary instead of permanent redirect during login

vgrichina:
When using 301 redirect it gets cached in Chrome and next time URL is opened server isn't even hit.

Fixes https://github.com/near/near-shell/issues/358

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,25 +77,32 @@ exports.login = async function (options) {
 
         // attempt to capture accountId automatically via browser callback
         let tempUrl;
-        // See: https://github.com/near/near-shell/issues/358
-        const isMac = process.platform === 'darwin';
+        const isWin = process.platform === 'win32';
 
         // find a callback URL on the local machine
         try {
-            // capture callback isn't working reliably outside of Mac. This is a workaround to not use it
-            if (isMac) {
+            if (!isWin) { // capture callback is currently not working on windows. This is a workaround to not use it
                 tempUrl = await capture.callback(5000);
-                // if we found a suitable URL, attempt to use it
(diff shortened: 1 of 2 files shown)


mikedotexe:
Wow, why does 302 behave so differently? 😵 
Worked for me.

index.js
```diff
@@ -77,27 +77,34 @@ exports.login = async function (options) {
...
-                await openUrl(newUrl);
             }
         } catch (error) {
             // console.error("Failed to find suitable port.", error.message)
             // TODO: Is it? Try triggering error
             // silent error is better here
         }
 
+        // if we found a suitable URL, attempt to use it
+        if (tempUrl) {
```
mikedotexe, line 94: we can move all this logic into the above `try catch` under `if (!isWin)` but I'm fine with it

vgrichina:
> Wow, why does 302 behave so differently? 😵

it kinda makes sense, one is temporary redirect, another is permanent redirect (so makes sense to avoid querying URL again).

index.js
vgrichina: @mikedotexe I just reverted previous change in this part of code.

mikedotexe:
I still don't understand why this works on OS X, then, but glad this is a solution.

vgrichina:
> alright, I'll revert the other PR that turned it off for Linux

hmm, this was already part of this PR

> I still don't understand why this works on OS X, then, but glad this is a solution.
I think it just works in Safari

vgrichina:
> I still don't understand why this works on OS X, then, but glad this is a solution.

Maybe it just worked in Safari? Safari doesn't cache permanent redirect it seems.

PR #364, opened 2020-05-11
Revert "address windows, mac, linux behavior w/ callback and auto-open Wallet"

vgrichina:
Reverts near/near-shell#361


mikedotexe:
I see you have this as draft, so I will hold off, but I am ready to revert this in favor of the 302 approach
https://github.com/near/near-shell/pull/363

PR #365, opened 2020-05-12
refactor: replace `nearlib` with `nearAPI`

### LINKED ISSUE #312: Replace nearlib with near-api-js
During the renaming process, we need to change shell's dependency from `nearlib` to`near-api-js`
Note: when this is complete, please make sure to change the `test_environment` and once that's done please alert folks publicly or internally for [the examples](https://github.com/near-examples). The examples are depending on the global object `nearlib` and we can rename that to `nearAPIJs`. We can expect example tests to fail momentarily after this change is merged.

chadoh:
Fixes #312 

We already switched dependencies from the `nearlib` library to `near-api-js`. This mostly completes the process by renaming the imported constant.

Given that `test_environment.js` is a dependency in use by many apps, I have opted to leave its initialization of a `nearlib` global in-place for now. Once we complete near-examples/counter#44, we will have a better idea of how to recommend people migrate away from `test_environment` altogether, at which point we can deprecate the whole file and link people to an upgrade guide. However, I have also opted to add a `nearAPI` global here. We can keep this unadvertised, but it seems strange now to set up a `nearlib` global but none for `nearAPI`

`near repl` is barely used by the current NEAR community, so I've updated the exported global name directly, without keeping the `nearlib` name around behind a deprecation warning.

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -1,5 +1,4 @@
-const nearlib = require('near-api-js');
-const { utils } = nearlib;
+const { providers, utils } = require('near-api-js');
 const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const inspectResponse = require('../utils/inspect-response');
@@ -34,7 +33,7 @@ async function scheduleFunctionCall(options) {
         JSON.parse(options.args || '{}'),
         options.gas,
         utils.format.parseNearAmount(options.amount));
-    const result = nearlib.providers.getTransactionLastResult(functionCallResponse);
+    const result = providers.getTransactionLastResult(functionCallResponse);
     console.log(inspectResponse(result));
     await eventtracking.track(eventtracking.EVENT_ID_SCHEDULE_FN_CALL_END, { node: options.nodeUrl, success: true });
 }
--- a/commands/repl.js
+++ b/commands/repl.js
@@ -5,7 +5,7 @@ module.exports = {
     handler: async (argv) => {
         const repl = require('repl');
         const context = repl.start('> ').context;
-        context.nearlib = require('near-api-js');
+        context.nearAPI = require('near-api-js');
         context.near = await require('../utils/connect')(argv);
         if (argv.accountId) {
             context.account = await context.near.account(argv.accountId);
--- a/test_environment.js
+++ b/test_environment.js
@@ -1,5 +1,5 @@
 const NodeEnvironment = require('jest-environment-node');
-const nearlib = require('near-api-js');
+const nearAPI = require('near-api-js');
 const fs = require('fs');
 
 const INITIAL_BALANCE = '500000000000000000000000000';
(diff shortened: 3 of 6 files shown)


chadoh:
Marking as draft because we don't want to merge with the `WHICH?` statements in place, but I would still like review. We can quickly replace these and merge once we have a firm idea of which version will remove deprecated `nearlib` references.

commands/repl.js
```diff
@@ -5,10 +5,16 @@ module.exports = {
     handler: async (argv) => {
         const repl = require('repl');
         const context = repl.start('> ').context;
-        context.nearlib = require('near-api-js');
+        context.nearAPI = require('near-api-js');
+        context.nearlib = new Proxy(context.nearAPI, {
```
vgrichina, line 9: let's just remove it, pretty sure nobody really depends on this yet (`near repl` has been broken for a while before anyone noticed)

test_environment.js
```diff
@@ -11,23 +11,29 @@ class LocalTestEnvironment extends NodeEnvironment {
     }
 
     async setup() {
-        this.global.nearlib = require('near-api-js');
+        this.global.nearAPI = require('near-api-js');
+        this.global.nearlib = new Proxy(this.global.nearAPI, {
+            get(obj, prop) {
+                console.warn('[DEPRECATION WARNING] `nearlib` has been deprecated and will be removed from near-shell@WHICH?; use `nearAPI` instead');
```
vgrichina, line 17: I think it's best to keep `nearlib` here, but issue deprecation for test_environment as a whole (i.e. change example tests to not depend on it).

I also think it makes sense to start using https://www.npmjs.com/package/depd for deprecation warnings as it has some UX improvements (like not spamming multiple times for same call-site).

commands/call.js
```diff
@@ -1,5 +1,5 @@
-const nearlib = require('near-api-js');
-const { utils } = nearlib;
+const nearAPI = require('near-api-js');
+const { utils } = nearAPI;
```
vgrichina, line 2: should we just change this to:

```js
const { utils, providers } = require('near-api-js');
```

I think it's preferable style to `nearAPI`

test_environment.js
chadoh: I like the idea of this, but before we deprecate the whole `test_environment` I want to be able to give people a clear upgrade path. In the deprecation message, we could link people to a PR in one of our example apps demonstrating how to remove the dependence on `test_environment`. 

I starting implementing such an upgrade in `token-contract-as`, but doing so is not straightforward yet. The simplest way to stop relying on `test_environment` is to copy a large chunk of it into your own app. I don't think we should rush removing `test_environment` until we have something better to recommend. I suspect that better way will be discovered as part of https://github.com/near-examples/counter/pull/44 

In the meantime, I think it's best to leave this file unchanged, so that we don't deprecate part of the `test_environment` interface in one version of `near-shell`, just to deprecate the entire thing a version or two later.

I've pushed an update with this change; please re-review the PR and leave a ✅ or ⛔️

vgrichina: > In the meantime, I think it's best to leave this file unchanged, so that we don't deprecate part of the test_environment interface in one version of near-shell, just to deprecate the entire thing a version or two later.

sgtm

Issue #314, opened 2020-04-16
inspect response should supply `maxArrayLength`


DanielRX:
If you run a getter that returns an array of length > 100 it will return `... x more items` instead of the full array, preventing any script from using the output of `near view` as it doesn't return the entire array

potatodepaulo:
Thanks for reporting. We're looking into this.

janedegtiareva:
Fix here https://github.com/near/near-shell/pull/366
This will be available soon in the next shell version.

PR #366, opened 2020-05-12
Print out full arrays, not just the first 100 elements

janedegtiareva:


janedegtiareva:
Fix for https://github.com/near/near-shell/issues/314

janedegtiareva:
Tested: see screenshot 
Screenshot (19)

janedegtiareva:
Tested on windows with a view function call that returns an array of 150 elements.

utils/inspect-response.js
```diff
@@ -1,5 +1,5 @@
 
 const util = require('util');
 module.exports = (response) => {
-    return util.inspect(response, { showHidden: false, depth: null, colors: true });
+    return util.inspect(response, { showHidden: true, depth: null, colors: true, maxArrayLength: null });
```
vgrichina, line 4: why is `showHidden` needed?

janedegtiareva:
> lgtm, but I think long term we need to expose some controls by user to make stuff more / less verbose

Yup, agreed.

utils/inspect-response.js
janedegtiareva: I found it useful since it printed out array length.

PR #370, opened 2020-05-16
Add support for using Ledger for signing and account creation

vgrichina:


middleware/ledger.js
```diff
@@ -0,0 +1,47 @@
+const { utils: { PublicKey, key_pair: { KeyType } } } = require('near-api-js');
+const { createClient } = require('near-ledger-js');
+const { default: TransportNodeHid } = require('@ledgerhq/hw-transport-node-hid');
+
+// near ... --useLedger --hdKeyPath
```
ilblackdragon, line 5: should this just be --useLedger=hd/key/path with default path by default?
also how about using account id (or hash of it) as key path?

mikedotexe:
This is great, I'm going to unbox my Ledger and try this out tonight

mikedotexe:
Hey @vgrichina I hate to be that person, but I think I need a little guidance on testing this PR. I'll hunt around for docs showing how to connect this to my Ledger, but if you have one that worked for you, could you share it?

middleware/ledger.js
vgrichina: > should this just be --useLedger=hd/key/path with default path by default?

Good catch, no need to have 2 args for this

> also how about using account id (or hash of it) as key path?
 
This is very good question and I am not sure what's best way to solve this. There are trade-offs involved. E.g. not being able to restore account easily when account name is forgotten. Like it still might doable (assuming there is a mode in Ledger app where you don't have to confirm getting public key), but you'll have to iterate through 2^24 combinations if we use all 3 bytes of path.

Also there is probability of collisions depending on how many bytes we use:
- 3 bytes: 99.97% probability of no collisions for 100 accounts
- 2 bytes: 92.72%
- 1 byte: close to 0%

I think more traditional way of using paths was keeping auto-increment counter somewhere, so that number of keys to iterate during recovery is smaller. This generally can even be Ledger itself (basically we can have API which is like "get new key"). We'll have to skip this for MVP though given app already in review, etc.

vgrichina:
@mikedotexe 

1. Follow README here to install app on Ledger:
https://github.com/near/near-ledger-app
2. Shell commands work as here:
https://github.com/near/near-shell/blob/1df2f9ce5bf160a8d7d688d4e2eee5a00f13bb45/middleware/ledger.js#L5

Note I'm likely to update it soon to `--useLedger=hd/key/path` as suggested by @ilblackdragon

middleware/ledger.js
vgrichina: > This generally can even be Ledger itself

To clarify on this: there is some amount of NVRAM we can use from app. This won't get recovered from seed phrase though, so account recovery still will have to iterate, by using auto-increment as strategy we are just reducing search scope.

Issue #306, opened 2020-04-03
Add usage info collection with devs permission


ilblackdragon:
We want to collect usage of the near shell to improve our products and figure out high level product metrics.

The flow should be post installation, we ask shell users if they would opt in to provide anonymized statistics of their usage. Can store their decision in `~/.neardev/config.json`.

If yes, we want to collect:
 - overall activations (e.g. summary of usage and session attribution?)
 - used networks (local vs devnet/betanet/testnet)
 - anything else? @potatodepaulo 

CC: @kcole16 for Amplitude

Also before fully rolling out, let's confirm with current developers that the language and usage info collected matches their expectations.

amgando:
resurfacing Gatsby as a best in class example

https://www.gatsbyjs.org/docs/telemetry/

janedegtiareva:
Do we want to add any more features? Can we close the issue at this point?

vgrichina:
> Do we want to add any more features? Can we close the issue at this point?

We've discussed in Slack that only interactive commands should ask for opt-in, fixed this here https://github.com/near/near-shell/pull/376

After digging into it I think I have better understanding of what we should add (in order of relative priorities):

- [ ] Provide command line option to opt-in / opt-out
- [ ] Include info on which platform event comes from. I mean GitPod vs CLI (not sure if OS is automatically tracked by mixpanel, but if not – we need it too).
- [ ] Report events even when there are errors (this is very interesting info for debugging at scale). Can be supplemented by some error tracking solution like http://getsentry.com.
- [ ] Connect session to web (e.g. when logging in to wallet)
- [ ] Connect sessions between different GitPod instances (not sure what's best way to approach this 
- [ ] Provide info necessary to connect with on-chain data (like accountId, etc). This might need discussion as doesn't really fall into `We only collect which commands are run via an anonymous identifier.`, though generally should be fine with opt-in (as long as user is informed).

I also have a concern around having to track both start and end events manually for every command. This seems pretty error prone long term (easy to forget to send end event, use wrong identifier while copy-pasting boilerplate, etc).

Is there any reason to not do it automatically for every command? Similar to how @kcole16  has implemented it in wallet https://github.com/near/near-wallet/blob/fb8f25ed17dbb49a662cbf3b6c10e3052b238827/src/middleware/index.js#L42 we just have a whitelist of Redux actions, but other than that they are logged automatically. Fr shell don't even need blacklist – Redux actions are just too noisy.

vgrichina:
[...]

PR #376, opened 2020-05-19
Ask for event tracking opt-in only in interactive commands

vgrichina:
For now this includes:
 - near login
 - near dev-deploy

Related to https://github.com/near/near-shell/issues/306

This change had to be done to make sure that opt-in flow doesn't interfere with usage in scripts.

### DIFF
--- a/utils/eventtracking.js
+++ b/utils/eventtracking.js
@@ -1,5 +1,5 @@
 const MIXPANEL_TOKEN = '9aa8926fbcb03eb5d6ce787b5e8fa6eb';
-var mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
+const mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
 
 const uuid = require('uuid');
 const chalk = require('chalk');  // colorize output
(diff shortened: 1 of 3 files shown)


utils/eventtracking.js
```diff
@@ -9,47 +9,14 @@ const settings = require('./settings');
 const TRACKING_ENABLED_KEY = 'trackingEnabled';
 const TRACKING_SESSION_ID_KEY = 'trackingSessionId';
 
+const shouldOptInByDefault = () => {
+    const isGitPod = !!process.env.GITPOD_WORKSPACE_URL;
```
janedegtiareva, line 13: Can we do just process.env.GITPOD_WORKSPACE_URL?

vgrichina: in this case we can, however it's pretty commonly used for explicit cast to boolean: https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript

main difference is if somebody downstream uses outside of logical operation (e.g. `shouldOptInByDefault() == = true` or `shouldOptInByDefault().toString()` – this can fail in unexpected way if not coerced into boolean here

I'd prefer to keep it as is because we return value outside of function (and it is reasonable for caller to expect it is boolean)

PR #377, opened 2020-05-19
Fix update notifier usage to support yarn

vgrichina:


mikedotexe:
I had a hard time understanding what this is doing, since I don't see `updateCommand` anywhere and there's no description here. So I hacked it so it'll output it, hardcoding version `9.1.0` and this is what I get:
Screenshot 2020-05-20 10 58 49

We don't want this, right? We're telling people to install `near-shell` globally with `npm` everywhere, but this will just add the latest version locally with `yarn` correct?

vgrichina:
@mikedotexe it is context-sensitive. If you install near-shell using npm – it'll output command for npm. You can test it like this:

1. Change `package.json` version to `0.23.2-foobar`
2. `npm i -g .`
3. you need to run some command 2 times (check is cached), then you'll get a banner:

```
   ╭──────────────────────────────────────────────────────────────────╮
   │                                                                  │
   │   NEAR Shell has a prerelease available 0.23.2-foobar → 0.23.2   │
   │    Run npm install -g near-shell to avoid unexpected behavior    │
   │                                                                  │
   ╰──────────────────────────────────────────────────────────────────╯
```

We hardcoded npm specifically for some reason which was giving incorrect message when installed with yarn, using local version from project, etc. This brings it back to how `update-notifier` package is designed.

vgrichina:
> Okay, I see it working as I'd hoped by doing:
> npm i near-shell@0.23.1 -g
> and then a couple times until it clicked:
> near state mike.testnet

this tests `near-shell@0.23.1`, not this PR

mikedotexe:
Jeez, I'm jumping around too much today and tested that in a silly way.
So I think this is still problematic @vgrichina 
```   ╭──────────────────────────────────────────────────────────╮
   │                                                          │
   │     NEAR Shell has a patch available 0.22.0 → 0.23.2     │
   │   Run yarn add near-shell to avoid unexpected behavior   │
   │                                                          │
   ╰──────────────────────────────────────────────────────────╯
```
But when I run this command:
```
yarn global v1.22.4
info "create-near-app@0.9.3" has binaries:
   - create-near-app
✨  Done in 0.17s.
```
it's not in the global for yarn. So this banner is essentially telling me to add yarn locally.
I'm using nvm btw.

PR #383, opened 2020-05-21
add delete-key command to delete access key

mikedotexe:
Delete's an access key.
Pull down branch and you can use this command
Command:
`./bin/near delete-key 2rLDDvyk8wdpq4TZ3k2xmCxhykq431eNKQ7Mfg31nT4q --accountId mike.testnet`

The access key can be copied from Wallet. Tested on full access and authorized apps keys.

### DIFF
--- a/index.js
+++ b/index.js
@@ -199,6 +199,16 @@ exports.sendMoney = async function (options) {
     await eventtracking.track(eventtracking.EVENT_ID_SEND_TOKENS_END, { node: options.nodeUrl, success: true });
 };
 
+exports.deleteAccessKey = async function (options) {
+    await eventtracking.track(eventtracking.EVENT_ID_DELETE_KEY_START, { node: options.nodeUrl, amount: options.amount });
+    console.log(`Deleting key = ${options.accessKey} on ${options.accountId}.`);
+    const near = await connect(options);
+    const account = await near.account(options.accountId);
+    const result = await account.deleteKey(options.accessKey);
+    console.log(inspectResponse(result));
+    await eventtracking.track(eventtracking.EVENT_ID_DELETE_KEY_END, { node: options.nodeUrl, success: true });
+};
+
 exports.stake = async function (options) {
     await eventtracking.track(eventtracking.EVENT_ID_STAKE_START, { node: options.nodeUrl, amount: options.amount });
     console.log(`Staking ${options.amount} (${utils.format.parseNearAmount(options.amount)}) on ${options.accountId} with public key = ${options.stakingKey}.`);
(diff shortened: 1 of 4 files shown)


bowenwang1996:
@mikedotexe why is this flag `--accessKey` needed?

mikedotexe:
@bowenwang1996 going to wait for tests to pass for posterity but then I'll merge and check with Vlad about cutting a release. If I don't hear back from him I'll cut one

mikedotexe:
> @mikedotexe why is this flag `--accessKey` needed?

you mean as opposed to just
`near delete-key mYHaSH… --accountId mike.testnet`?

index.js
```diff
@@ -199,6 +199,16 @@ exports.sendMoney = async function (options) {
     await eventtracking.track(eventtracking.EVENT_ID_SEND_TOKENS_END, { node: options.nodeUrl, success: true });
 };
 
+exports.deleteAccessKey = async function (options) {
```
vgrichina, line 202: nit: I'd prefer we add new commands as separate self-contained files (i.e. in `commands/` folder). We should extract existing commands too eventually.

mikedotexe: okay, pushed up that change, Vlad

utils/eventtracking.js
```diff
@@ -109,5 +109,7 @@ module.exports = {
     EVENT_ID_REPL_END: 'shell_repl_end',
     EVENT_ID_GENERATE_KEY_START: 'shell_generate_key_start',
     EVENT_ID_GENERATE_KEY_END: 'shell_id_generate_key_end',
+    EVENT_ID_DELETE_KEY_START: 'event_id_delete_key_start',
```
vgrichina, line 112: @mattlockyer not really comment on this PR, but relevant context – what do you think on https://github.com/near/near-shell/issues/306#issuecomment-630612739

Should we make events automatic for commands?

mattlockyer: We're trying to capture the best metrics. We won't get it right the first time. We should be opt-in and collect everything we can, unless the user explicitly opts out.
Therefore a whitelist for events seems like an anti-pattern and all events should get logged automatically with a blacklist (if things are too noisy). But noise is also something you can just filter out later in your analytics.

vgrichina: just realized I wanted to tag @mikedotexe (as it's his PR), but thanks for input! I agree that logging by default is good idea for near-shell commands

mikedotexe: @vgrichina I think I'll answer over there in the issue. I really don't like the START and END, but we haven't yet implemented crash reporting so doing what we have. I think the more events we track the better. Thanks for tagging that other issue, I'll provide feedback

Issue #393, opened 2020-06-03
near login --help is not correct


nearmax:
near version is `0.24.1`.

Here is what `near login --help` returns:
```
near login

logging in through NEAR protocol wallet

Options:
  --help                     Show help  [boolean]
  --version                  Show version number  [boolean]
  --nodeUrl, --node_url      NEAR node URL  [string] [default: "http://localhost:3030"]
  --networkId, --network_id  NEAR network ID, allows using different keys based on network  [string] [default: "default"]
  --helperUrl                NEAR contract helper URL  [string]
  --keyPath                  Path to master account key  [string]
  --accountId, --account_id  Unique identifier for the account  [string]
  --useLedgerKey             Use Ledger for signing with given HD key path  [string] [default: "44'/397'/0'/0'/1'"]
```
Note, that it says that by the default it connects to the local network. It, however, connects to the testnet.

PR #395, opened 2020-06-03
Fix https://github.com/near/near-shell/issues/393

june07:
Fixed help to match what is actually happening.

node bin\near login --help
near login

logging in through NEAR protocol wallet

Options:
  --help                     Show help  [boolean]
  --version                  Show version number  [boolean]
  --nodeUrl, --node_url      NEAR node URL  [string] [default: "https://rpc.testnet.near.org"]
  --networkId, --network_id  NEAR network ID, allows using different keys based on network  [string] [default: "default"]
  --helperUrl                NEAR contract helper URL  [string]
  --keyPath                  Path to master account key  [string]
  --accountId, --account_id  Unique identifier for the account  [string]
  --useLedgerKey             Use Ledger for signing with given HD key path  [string] [default: "44'/397'/0'/0'/1'"]

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -140,12 +140,12 @@ yargs // eslint-disable-line
     .option('nodeUrl', {
         desc: 'NEAR node URL',
         type: 'string',
-        default: 'https://rpc.testnet.near.org'
+        default: config.nodeUrl
     })
     .option('networkId', {
         desc: 'NEAR network ID, allows using different keys based on network',
         type: 'string',
-        default: 'default'
+        default: config.networkId
     })
     .option('helperUrl', {
         desc: 'NEAR contract helper URL',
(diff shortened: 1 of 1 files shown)


bin/near-cli.js
```diff
@@ -140,7 +140,7 @@ yargs // eslint-disable-line
     .option('nodeUrl', {
         desc: 'NEAR node URL',
         type: 'string',
-        default: 'http://localhost:3030'
+        default: 'https://rpc.testnet.near.org'
```
chadoh, line 143: Looks like this is coming from `config.js`, and the default environment is `development`

Looks like `config` is defined on line 135 above. Could we use it here, to make sure we always give accurate information?

```js
default: option.nodeUrl
```

We could also update the `.option('networkId'` default while we're here.

This means that if you used `NEAR_ENV=devnet near login --help`, it would show the devnet values in the help message. I think that seems like what we would want. Do you agree?

PR #401, opened 2020-06-05
Add basic crash reporting to near-cli

janedegtiareva:
This requires adding a workaround to spawn a child process, since
mixpanel does not handle this case properly.

There is an existing limitation that if a user did not opt in to
tracking, the crash report will not be sent. We may want to ask an
additional qustion for this case, since there are scenarios where people
are willing to send an individual crash report but not opt in to general
tracking.

There is a known limitation that this implementation does not log the
environment that was used.

### DIFF
--- a/utils/crash-error-report.js
+++ b/utils/crash-error-report.js
@@ -5,5 +5,4 @@ const mixPanelProperties = {
     error_message: process.env.NEAR_CLI_LAST_ERROR,
     network_id: process.env.NEAR_CLI_NETWORK_ID
 };
-Object.assign(mixPanelProperties, {});
 eventtracking.track(eventtracking.EVENT_ID_ERROR, mixPanelProperties);
--- a/utils/crash-error-report.js
+++ b/utils/crash-error-report.js
@@ -3,7 +3,7 @@ const eventtracking = require('./eventtracking');
 const mixPanelProperties = {
   near_shell_command: process.env.NEAR_CLI_ERROR_LAST_COMMAND,
   error_message: process.env.NEAR_CLI_LAST_ERROR,
-  environment: process.env.NODE_ENV 
+  network_id: process.env.NEAR_CLI_NETWORK_ID
 };
 Object.assign(mixPanelProperties, {});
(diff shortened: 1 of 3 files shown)


janedegtiareva:
Tested on windows, ` node bin\near state cad10` produces this in mixpanel 
Screenshot (27)

janedegtiareva:
There is a known issue that if a user fires off 2 commands in a row that result in a crash, only one gets logged. I haven't investigated the root cause. Possible explanations: some sort of event-deduping in mixpanel, process forking.

vgrichina:
Should we integrate with http://getsentry.com to get more context for errors (stack trace, etc)?

janedegtiareva:
> Should we integrate with http://getsentry.com to get more context for errors (stack trace, etc)?

That's a good option for the next iteration @ashleynear @potatodepaulo . I think we'll learn a lot from this one as a easy-to-implement low hanging fruit, so it still makes sense to land this.

janedegtiareva:
This PR was originally based on this POC work by @mikedotexe https://github.com/near/near-shell/commit/e1c5754c10b812217d5815ee5578cef78be4b5a2

utils/crash-error-report.js
```diff
@@ -0,0 +1,9 @@
+const eventtracking = require('./eventtracking');
+
+const mixPanelProperties = {
+    near_shell_command: process.env.NEAR_CLI_ERROR_LAST_COMMAND,
+    error_message: process.env.NEAR_CLI_LAST_ERROR,
+    network_id: process.env.NEAR_CLI_NETWORK_ID
+};
+Object.assign(mixPanelProperties, {});
```
mikedotexe, line 8: I don't understand this line, isn't this adding nothing to the `mixPanelProperties`?
Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

mikedotexe: tested this by removing this line and it works, I think we can get rid of it

janedegtiareva: This was a copy paste artifact, thanks for catching!

PR #402, opened 2020-06-05
In create_account, create key first to avoid unusable accounts and funds

mikedotexe:
Two things to address here. 
1. DevX usability as we frequently see timeout errors
2. I believe we need this long term, especially for mainnet.

Regarding 2, take this example. A user creates a subaccount on mainnet, sending it a lot of valuable tokens with
`near create_account rich.mike.near --masterAccount mike.near --initialBalance 999999`
Currently the RPC request is called *first* with `await` and if successful, the key file is created.
If lightning strikes this person's house and the internet goes out **immediately after the RPC call** and before the key file is created, this account is bricked and the user has lost 999999 Ⓝ.

What this PR does is create the key file first, then make the async calls. If there's an error it'll delete the key file. Caveat: if that error is a timeout, it'll prompt the user with useful information and *not* delete the key file.

I tried using the error's `type` but sadly that wont work. I have to parse the error message for `Timeout`. The `type = UntypedError`. It's always like this:

```
Error:  TypedError: [-32000] Server error: Timeout error
    at JsonRpcProvider.sendJsonRpc (/Users/mike/.nvm/versions/node/v12.16.0/lib/node_modules/near-shell/node_modules/near-api-js/lib/providers/json-rpc-provider.js:129:27)
    at async /Users/mike/.nvm/versions/node/v12.16.0/lib/node_modules/near-shell/commands/tx-status.js:35:24
    at async Object.handler (/Users/mike/.nvm/versions/node/v12.16.0/lib/node_modules/near-shell/utils/exit-on-error.js:4:9) {
  type: 'UntypedError'
}
```

Here's a patch file (I had to add the `.txt` extension so that Github would upload it) for `nearcore` to hack it to return a timeout error:
[always-server-timeout.patch.txt](https://github.com/near/near-shell/files/4737192/always-server-timeout.patch.txt)

Note: I wasn't able to use `await` in the `catch` hence the use of `.then` in this PR.

I've included a video of it in action. First we show that the file is being deleted when parsing the error message for an error that's *note* `Timeout` to illustrate that it deletes the key file afterward. Then we fix it to look for `Timeout` and observe that the key file will persist and the message is displayed to the user.

Video link from Discord: https://discordapp.com/channels/490367152054992913/542945453533036544/718492520320991302

### DIFF
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -55,13 +55,13 @@ async function createAccount(options) {
         await near.createAccount(options.accountId, publicKey);
     } catch(error) {
         if (error.message.includes('Timeout')) {
-            console.warn("Received a timeout when creating account, please run:");
+            console.warn('Received a timeout when creating account, please run:');
             console.warn(`near state ${options.accountId}`);
-            console.warn("to confirm creation. Keyfile for this account has been saved.");
+            console.warn('to confirm creation. Keyfile for this account has been saved.');
         } else {
             near.connection.signer.keyStore.removeKey(options.networkId, options.accountId)
                 .then(() => {
-                    throw error
+                    throw error;
                 }).catch((e) => console.error(e));
         }
     }
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -131,19 +131,17 @@ async function createAccount(options) {
     // Create account
     try {
         await near.createAccount(options.accountId, publicKey);
+        console.log(`Account ${options.accountId} for network "${options.networkId}" was created.`);
     } catch(error) {
         if (error.message.includes('Timeout')) {
             console.warn('Received a timeout when creating account, please run:');
             console.warn(`near state ${options.accountId}`);
             console.warn('to confirm creation. Keyfile for this account has been saved.');
         } else {
-            near.connection.signer.keyStore.removeKey(options.networkId, options.accountId)
-                .then(() => {
-                    throw error;
-                }).catch((e) => console.error(e));
+            await near.connection.signer.keyStore.removeKey(options.networkId, options.accountId)
+            throw error;
         }
     }
-    console.log(`Account ${options.accountId} for network "${options.networkId}" was created.`);
 }
 
 module.exports = {
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -81,18 +81,16 @@ async function createAccount(options) {
     if (splitAccount.length === 1) {
         // TLA (bob-with-at-least-maximum-characters)
         if (splitAccount[0].length < TLA_MIN_LENGTH) {
-            console.log(`Top-level accounts must be at least ${TLA_MIN_LENGTH} characters.\n` +
+            throw new Error(`Top-level accounts must be at least ${TLA_MIN_LENGTH} characters.\n` +
               'Note: this is for advanced usage only. Typical account names are of the form:\n' +
               'app.alice.test, where the masterAccount shares the top-level account (.test).'
             );
-            return;
         }
     } else if (splitAccount.length > 1) {
         // Subaccounts (short.alice.near, even.more.bob.test, and eventually peter.potato)
         // Check that master account TLA matches
         if (!options.accountId.endsWith(`.${options.masterAccount}`)) {
-            console.log(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
-            return;
+            throw new Error(`New account doesn't share the same top-level account. Expecting account name to end in ".${options.masterAccount}"`);
         }
 
         // Warn user if account seems to be using wrong network, where TLA is captured in config
@@ -121,8 +119,7 @@ async function createAccount(options) {
     try {
         // This is expected to error because the account shouldn't exist
         await near.account(options.accountId);
-        console.error(`Sorry, account '${options.accountId}' already exists.`);
-        process.exit(1);
+        throw new Error(`Sorry, account '${options.accountId}' already exists.`);
     } catch (e) {
         if (!e.message.includes('does not exist while viewing')) {
             throw e;
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -117,6 +117,19 @@ async function createAccount(options) {
     } else {
         await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_END, { success: true, new_keypair: false }, options);
     }
+    // Check to see if account already exists
+    try {
+        // This is expected to error because the account shouldn't exist
+        await near.account(options.accountId);
+        console.error(`Sorry, account '${options.accountId}' already exists.`);
+        return;
+    } catch (e) {
+        if (!e.message.includes('does not exist while viewing')) {
+            console.error('Error while checking for existence of account', e);
+            return;
+        }
+    }
+    // Create account
     try {
         await near.createAccount(options.accountId, publicKey);
     } catch(error) {
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -130,7 +130,7 @@ async function createAccount(options) {
         await near.createAccount(options.accountId, publicKey);
         console.log(`Account ${options.accountId} for network "${options.networkId}" was created.`);
     } catch(error) {
-        if (error.message.includes('Timeout')) {
+        if (error.type === 'RetriesExceeded') {
             console.warn('Received a timeout when creating account, please run:');
             console.warn(`near state ${options.accountId}`);
             console.warn('to confirm creation. Keyfile for this account has been saved.');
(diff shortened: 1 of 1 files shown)


commands/create-account.js
```diff
@@ -48,10 +48,23 @@ async function createAccount(options) {
...
         publicKey = keyPair.getPublicKey();
     }
-    await near.createAccount(options.accountId, publicKey);
     if (keyPair) {
         await near.connection.signer.keyStore.setKey(options.networkId, options.accountId, keyPair);
     }
+    try {
+        await near.createAccount(options.accountId, publicKey);
+    } catch(error) {
+        if (error.message.includes('Timeout')) {
```
janedegtiareva, line 57: some food for thought for future work: how can we extend this so that we handle timeouts better across the board?

mikedotexe: I believe this is deep in the chain, and the solution is to handle it there.

commands/create-account.js
```diff
@@ -48,10 +48,23 @@ async function createAccount(options) {
...
+            console.warn('Received a timeout when creating account, please run:');
+            console.warn(`near state ${options.accountId}`);
+            console.warn('to confirm creation. Keyfile for this account has been saved.');
+        } else {
+            near.connection.signer.keyStore.removeKey(options.networkId, options.accountId)
+                .then(() => {
```
vgrichina, line 63: @mikedotexe avoid mixing `await` and `.then` or you end up having unexpected results (eg. here error won't propagate properly to return value of this function.

commands/create-account.js
vgrichina, line 57: I think we already should have `type` field on `error` to avoid searching  string in message.

mikedotexe: I don't understand why, but I tried using `await` and it didn't work.

mikedotexe: I wish, the `type` is `UntypedError` which I really don't understand. In the PR description I included a patch to hack nearcore to always return Server Timeout and you can see this behavior.

commands/create-account.js
mikedotexe: I don't understand why, but I tried using await and it didn't work.

commands/create-account.js
vgrichina: I think included nearcore patch is not returning error in same way as it is normally returned. 

Normally it is:

```rust
.map_err(|err| RpcError::server_error(Some(ServerError::from(err))))
```
i.e. it is converted/wrapped into `RpcError::server_error` and otherwise returned as `Ok` response

in your patch it is:

```rust
Err(timeout_err())
```

- using `timeout_err` directly without conversion/wrapping
- using `Err` instead of `Ok`

mikedotexe: I just had a couple timeouts trying to create a new account (now those 2 accounts are bricked, no way to recover the key)
The errors are the same.

Screenshot 2020-06-09 15 02 20

vgrichina: @bowenwang1996 @fckt do you have any idea what is going on here? Shouldn't nearcore already return typed errors in every case?

bowenwang1996: I am very sorry but looks like it still hasn't been fixed. We scammed you :(

commands/create-account.js
vgrichina: May you give more context what did not work? 

Current solution has some issues:
 - code doesn't wait until `removeKey` call completion
 - error gets ignored after logging (maybe that's intentional, but other errors usually would result in non-zero return code for shell)

It seems like it should work like this:
```js
await near.connection.signer.keyStore.removeKey(options.networkId, options.accountId);
throw error;
```

or if the intent is to swallow `removeKey` error like this:

```js
try {
    await near.connection.signer.keyStore.removeKey(options.networkId, options.accountId);
} catch(innerError) {
    console.error(`Error removing key for ${options.accountId}`, innerError);
} finally {
    throw error;
}

```


Let me know if you need help debugging this – let's figure it out together. Feel free to ping me next time if you have any questions regarding `await` / promises.

commands/create-account.js
vgrichina: @mikedotexe looks like we have to roll with `error.message.includes` for a while then

mikedotexe:
> @mikedotexe just realized that trying to create account twice would remove the key altogether with current implementation

Thanks for that, Vlad. That's a mistake on my part. This is on my radar, although I'll be jumping on other pressing tasks. Wanted to have some follow-up. This is an important enhancement, but I will be getting to it as time allows.

commands/create-account.js
janedegtiareva: I agree, let's fix this.

mikedotexe:
> Looks good after addressing the await vs .then comment.

hey @janedegtiareva I showed Vlad during our one-on-one that for some crazy reason, the `.then` is necessary and `await` doesn't work. I wonder if it's something about it being in the scope of a `try catch` but am not sure. I actually got interested enough to hack at another part of near-shell which is the whole crash reporting side of Mixpanel. I thought maybe the same approach would allow us to directly call Mixpanel instead of the fork technique, but no such luck. So anyway, for whatever reason, it seems like we need `.then`. I'm just chalking it up to yet another "JavaScript is weird" list in my head and moving on. :)

mikedotexe:
We just had an end user lost tokens because of a timeout.
Adding this for tracking purposes, I suppose:

Screenshot 2020-06-24 16 18 02

commands/create-account.js
```diff
@@ -117,6 +117,19 @@ async function createAccount(options) {
     } else {
         await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_END, { success: true, new_keypair: false }, options);
     }
+    // Check to see if account already exists
+    try {
+        // This is expected to error because the account shouldn't exist
+        await near.account(options.accountId);
+        console.error(`Sorry, account '${options.accountId}' already exists.`);
+        return;
```
vgrichina, line 125: should return failure exit code, this returns success

commands/create-account.js
```diff
@@ -117,6 +117,19 @@ async function createAccount(options) {
...
+    } catch (e) {
+        if (!e.message.includes('does not exist while viewing')) {
+            console.error('Error while checking for existence of account', e);
```
vgrichina, line 128: it's best to just throw error from here (otherwise success exit code is returned)

commands/create-account.js
vgrichina: Looks like after near-api-js fix we would get `RetriesExceeded` error:

https://github.com/near/near-api-js/blob/04f7c2de6e59545f6832ade0f40f608a5c7d41e9/src/account.ts#L132

commands/create-account.js
vgrichina: @mikedotexe this is not fixed

commands/create-account.js
mikedotexe: pushed, this was an oversight on my part, whoops. thanks for catching

commands/create-account.js
mikedotexe: pushed as well

commands/create-account.js
vgrichina: nit: probably makes sense to throw vs call `process.exit` directly (to be consistent with other errors)

vgrichina:
https://github.com/near/near-shell/pull/402#discussion_r437127111 looks like this not fixed yet @mikedotexe

vgrichina:
@mikedotexe also please mark this as ready for review when done, as otherwise it looks like doesn't show up anywhere without going to specific PR

commands/create-account.js
mikedotexe: > [#402 (comment)](https://github.com/near/near-shell/pull/402#discussion_r437127111) looks like this not fixed yet @mikedotexe

@vgrichina we talked about this during our 1-on-1, remember? I shared my screen and showed you that `await` and `.then` behaved differently for some unknown reason.

commands/create-account.js
mikedotexe: there isn't an actual error here, though, right? This is in the `try` section.

Screenshot 2020-07-10 15 11 54

commands/create-account.js
vgrichina: Now I remember that we talked about this, but I don't exactly remember what was the problem already – so good idea to add a comment. Looks like I have in the notes that key didn't get removed when using `.removeKey`, is this correct? I'll try to reproduce it and fix.

Even if we have to use `.then` there are other problems:
1) Looks like error won't propagate and process will return success code even though command failed.
2) Seems that `console.log(`Account ${options.accountId} for network "${options.networkId}" was created.`);` will print the message anyway, even though account wasn't created successfully.

commands/create-account.js
vgrichina: what I mean is that we can do:

```
throw new Error(`Sorry, account '${options.accountId}' already exists.`)
```

so that we hit normal error handling path. It's not like super important by itself but makes stuff easier if we want to do some common stuff on error (like report it to analytics).

vgrichina:
https://github.com/near/near-shell/pull/402/files#r452670751 @mikedotexe see this as currently you won't get Timeout error (and so PR won't solve the problem it is trying to solve).

mikedotexe:
Thanks for the commits, Vlad

PR #403, opened 2020-06-05
Add data on whether we are in gitpod. Use gitpod user id hash for id

janedegtiareva:


janedegtiareva:
Fix for https://app.zenhub.com/workspace/o/near/devx/issues/194

janedegtiareva:
Fix for https://github.com/near/devx/issues/194

janedegtiareva:
Tested on windows local + gitpod.
*gitpod (see screenshot)
*local with missing consent data in settings (question hasn't been asked)
*local with user opt out of tracking
*local with user opt in to tracking 
Screenshot (29)

utils/eventtracking.js
```diff
@@ -3,26 +3,44 @@ const mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
...
 
+const isGitPod = () => {
+    return !!process.env.GITPOD_WORKSPACE_URL;
+};
+
+const getGitPodUserHash = () => {
+    if (!process.env.GITPOD_GIT_USER_NAME) {
+        return null;
+    }
+    return crypto.createHash('sha256').update(process.env.GITPOD_GIT_USER_NAME, 'utf8').digest('hex').toString();
```
vgrichina, line 21: Cool that we can track sessions across users in anonymized way. Wonder if it's better to use `GITPOD_GIT_USER_EMAIL` as my understanding is that `GITPOD_GIT_USER_NAME` is full name (so not necessarily unique).

janedegtiareva: sure, updated.

PR #404, opened 2020-06-05
Track network id and node url consistently

janedegtiareva:
This PR refactors the way we track node url to be more consistent by using data from options instead of asking the caller to pass it.

This PR also contains an example of how we might want to add additional tracking to specific user flows to be to track user journeys. (see login)

Potential future work: log an event on invoking shell functions automatically

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -22,7 +22,7 @@ module.exports = {
 };
 
 async function scheduleFunctionCall(options) {
-    await eventtracking.track(eventtracking.EVENT_ID_SCHEDULE_FN_CALL_START, { node: options.nodeUrl });
+    await eventtracking.track(eventtracking.EVENT_ID_SCHEDULE_FN_CALL_START, {}, options);
     console.log(`Scheduling a call: ${options.contractName}.${options.methodName}(${options.args || ''})` +
         (options.amount && options.amount != '0' ? ` with attached ${options.amount} NEAR` : ''));
     const near = await connect(options);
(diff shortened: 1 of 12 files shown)


janedegtiareva:
Tested: windows command line and gitpod

janedegtiareva:
> I think that since we merged the crash reporting, we'll want to remove all the EVENT_…_END logging, right? And in that case, we can remove the `{ success: true },` from all of them, and only capture the:
> `const track = async (eventType, options) => {`
> unless we want to do this in a separate PR, but this would be a reasonable place/time to do it I believe.
> works great over here with testing

This is not a slam-dunk decision. Check login for example - there are user journey paths that we want to detect (they came back to shell, but account key verification failed) which will not get caught by the crash reporting code.

vgrichina:
> This is not a slam-dunk decision. Check login for example - there are user journey paths that we want to detect (they came back to shell, but account key verification failed) which will not get caught by the crash reporting code.

yeah, I agree that we still would need to log some events manually. 

I think in ideal world we do:
 - log start event automatically for every command
 - log end event with/without error automatically for every command
 - log various non-error special cases manually (error cases can just have necessary context in error thrown)

janedegtiareva:
I moved the code that tracks command invocation to exitOnError, so that it's co-located with crash reporting code. Note that the same workaround with starting a new process needs to be used -- in case the shell crashes quickly, the command invocation event will not get logged unless it's done that way. 
I removed the non-trivial *_END events and left the ones that carry useful information. 

There are a few known small issues: 
* If shell crashes instantaneously, the ordering of the _START event and the crash report event in mixpanel is non-deterministic (sometimes the crash report event appears before the corresponding command fired event)
* repl does not produce tracking events
* if a user needs to be asked to opt in to tracking, the _START event for the command where this happened will not fire (at the time of logging this, the consent has not been given yet)

PR #405, opened 2020-06-05
Make case consistent in near command - fix near create_account

janedegtiareva:
* add create-account command
* create_account still works, but logs a deprecation warning. It does not show up in near help

Tested: with error handing 
```
C:\Users\janed\near\near-shell>node bin\near create-account  n --masterAccount a
Using options: {
  masterAccount: 'a',
  networkId: 'default',
  nodeUrl: 'https://rpc.testnet.near.org',
  contractName: undefined,
  walletUrl: 'https://wallet.testnet.near.org',
  helperUrl: 'https://helper.testnet.near.org',
  useLedgerKey: "44'/397'/0'/0'/1'",
  newLedgerKey: "44'/397'/0'/0'/1'",
  initialBalance: '100000000000000000000000000',
  accountId: 'n'
}
Error:  TypedError: [-32000] Server error: Account ID 'a' is not valid
    at JsonRpcProvider.sendJsonRpc (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\providers\json-rpc-provider.js:129:27)
    at async JsonRpcProvider.query (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\providers\json-rpc-provider.js:61:24)
    at async Account.fetchState (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\account.js:46:23)
    at async Account.signAndSendTransaction (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\account.js:87:9)
    at async LocalAccountCreator.createAccount (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\account_creator.js:24:9)
    at async Near.createAccount (C:\Users\janed\near\near-shell\node_modules\near-api-js\lib\near.js:53:9)
    at async createAccount (C:\Users\janed\near\near-shell\commands\create-account.js:84:5)
    at async Object.handler (C:\Users\janed\near\near-shell\utils\exit-on-error.js:4:9) {
  type: 'UntypedError'
}

### CODE REVIEW

**Comment by janedegtiareva** [comment]
This is one of the work items needed to complete https://github.com/near/devx/issues/203

File: bin/near-cli.js [hunk 1]
```diff
@@ -184,7 +184,8 @@ yargs // eslint-disable-line
     .middleware(require('../middleware/print-options'))
     .middleware(require('../middleware/key-store'))
     .middleware(require('../middleware/ledger'))
-    .command(require('../commands/create-account'))
+    .command(require('../commands/create-account').createAccountCommand)
+    .command(require('../commands/create-account').createAccountCommandDeprecated)
```
chadoh, line 188: I'd rather stick with the convention in this file of using the default export. Now that we have two, I would expect the default export to be the correct one. Maybe we could export the deprecated one as a named export, that way when we remove it, the interface will be parallel with all the others here. That would make this file look like this:

```js
    .command(require('../commands/create-account'))
    .command(require('../commands/create-account').deprecated)
```

commands/create-account.js
```diff
@@ -36,6 +36,39 @@ module.exports = {
     handler: exitOnError(createAccount)
 };
 
+const createAccountCommandDeprecated = {
```
chadoh, line 39: Are we able to use `import/export` syntax? If so, I'd prefer we put it inline

```js
export const deprecated = {

File: commands/create-account.js [hunk 3]
```diff
@@ -4,8 +4,8 @@ const connect = require('../utils/connect');
 const { KeyPair } = require('near-api-js');
 const eventtracking = require('../utils/eventtracking');
 
-module.exports = {
-    command: 'create_account <accountId>',
+const createAccountCommand = {
+    command: 'create-account <accountId>',
     desc: 'create a new developer account',
     builder: (yargs) => yargs
         .option('accountId', {
```
chadoh, line 11: Don't we want to change these options to use dash-case also? I think `yargs` will accept them in either format, but I think we should probably show them in dash-case in any help commands.

commands/create-account.js
```diff
@@ -36,6 +36,39 @@ module.exports = {
...
+            type: 'string',
+            default: "44'/397'/0'/0'/1'"
+        })
+        .option('initialBalance', {
+            desc: 'Number of tokens to transfer to newly created account',
+            type: 'string',
+            default: '100'
+        }),
+    handler: exitOnError(async (options) => {
+        console.log('near create_account is deprecated. Please use near create-account.');
```
chadoh, line 68: It might be nice if we say in which version of near-shell we plan to remove the old version

commands/create-account.js
janedegtiareva: fixing options is a larger undertaking which has some risk (see https://github.com/near/near-shell/pull/414), so I'd prefer to do it as a separate effort.

chadoh: Yeah, I wondered if that would be more work than it's worth right now. I think the change from `create_account` to `create-account` is important enough that we can leave the options as-is for now.

bin/near-cli.js
janedegtiareva: It's probably not worth the while to spend time on this for temporary code.

commands/create-account.js
janedegtiareva: Good idea! Let's give it 2 minor version bumps

janedegtiareva:
> See some ideas for further improvements below, but I'd be ok merging this as-is. Thanks!

Took some suggestions and punted on others. Thanks!

mikedotexe:
I just realized there are tests that use `create_account` and we should fix those

Issue #164, opened 2019-10-22
Invalid key shows weird error


ilblackdragon:
They key contains %3 from copying and this gets non-base58 error
```
near stake sparkpool ed25519%3ACSUZgqBHtpUyycqRmZe8yg1QgERpc7MzmihGS7kB8zfw 100000000000
[WARNING] Didn't find config at /root/src/config, using default shell config
Staking 100000000000 on sparkpool with public key = ed25519%3ACSUZgqBHtpUyycqRmZe8yg1QgERpc7MzmihGS7kB8zfw.
Error:  Error: Non-base58 character
    at Object.decode (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/node_modules/base-x/src/index.js:112:11)
    at Object.base_decode (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/node_modules/nearlib/lib/utils/serialize.js:16:39)
    at Function.fromString (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/node_modules/nearlib/lib/utils/key_pair.js:42:63)
    at Function.from (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/node_modules/nearlib/lib/utils/key_pair.js:35:30)
    at Account.stake (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/node_modules/nearlib/lib/account.js:133:110)
    at Object.exports.stake (/root/.nvm/versions/node/v10.0.0/lib/node_modules/near-shell/index.js:139:34)
    at process._tickCallback (internal/process/next_tick.js:178:7)
```

icerove:
our public key is (base58 encoded), so error is right

june07:
Seems to me the issue is the "%3A" which is the url encoded value of ":" being input into the near-api-js library... which doesn't seem to decode the key string before processing.

Assuming the near-api-js library is the standard and correct way of doing things, we should be url decoding the user input.  PR issues for fix https://github.com/near/near-shell/pull/406.

PR #406, opened 2020-06-07
Fix for #164

june07:
Tested locally by running the command as per #164 and got no error.  Further the URL decoded staking key is also printed to the console.


mikedotexe:
Thanks for contributing! I look forward to taking a peek soon

janedegtiareva:
Would you mind re-basing to latest? We had an issue on master which caused build failures, and it's now fixed. 
Thank you!

PR #414, opened 2020-06-08
Ledger implementation does not interact well with option aliases - repro

janedegtiareva:


janedegtiareva:
This change breaks view state: 
`node bin\near state test.near` now produces `Make sure to connect your Ledger and open NEAR app`

janedegtiareva:
What happens is that in the code, shell uses a camelCase convention for option values throughout. Most of time times aliases work because aliasing options results in all the possible cases being set in the options object which is being passed around. However, this is an example of usage which refers to some other entity which does not get stuff set according to aliases

middleware/ledger.js
```diff
@@ -3,6 +3,7 @@ const { utils: { PublicKey, key_pair: { KeyType } } } = require('near-api-js');
 // near ... --useLedgerKey
 // near create_account new_account_name --newLedgerKey --useLedgerKey --masterAccount account_that_creates
 module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, newLedgerKey, publicKey }, yargs) {
+    console.log(yargs)
     if (yargs.parsed.defaulted.useLedgerKey) {
```
mikedotexe, line 7: I think we need to change this to

    if (yargs.parsed.defaulted['use-ledger-key']) {

janedegtiareva: Correct - that's the way to fix this particular issue. 
The problem that I am trying to demonstrate is that we need to do a regression test pass on a lot of shell functionality if we proceed with the naming convention change (i.e. this is no longer a low-hanging-fruit)

mikedotexe: Actually, I'm not sure this is a deeper problem. Or I still don't understand what this PR is made to imply. I think what happened was we told `near-shell` there's a new option that has a couple aliases, and then we referred to a key that no longer exists because we changed it.
Maybe I don't understand something here

janedegtiareva: It highlights the potential regression impact on switching to dash-case as the canonical option form. This is fine as long as we are willing to re-test all the commands when we do this!

PR #418, opened 2020-06-09
Use new mixpanel token so that we route events into Shell & wallet proj

janedegtiareva:


janedegtiareva:
https://github.com/near/devx/issues/191

janedegtiareva:
Tested: on windows, see screenshot with events populating in the new project. 
Screenshot (31)

PR #440, opened 2020-07-01
Add verbose flag which can be used to control amount of logs

janedegtiareva:
* Do not print out connection options if verbose is off.

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -183,7 +183,8 @@ yargs // eslint-disable-line
     .option('verbose', {
         desc: 'Prints out verbose output',
         type: 'boolean',
-        default: true
+        alias: 'v',
+        default: false
     })
     .middleware(require('../middleware/initial-balance'))
     .middleware(require('../middleware/print-options'))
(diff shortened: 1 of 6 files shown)


bin/near-cli.js
```diff
@@ -180,6 +180,11 @@ yargs // eslint-disable-line
         desc: 'Expected top-level account for a network',
         type: 'string'
     })
+    .option('verbose', {
+        desc: 'Prints out verbose output',
+        type: 'boolean',
+        default: true
```
chadoh, line 186: I'm strongly in favor of making this `default: false`. That's much more standard in CLIs – by default, you don't get inundated with output. You have to add `--verbose` or `-v` for that.

Right now, this will be backwards – inundation by default, and you have to type what? `--verbose=false`? in order to turn it off? I don't like it.

Let's also add the `-v` alias, please!

mikedotexe: While I agree that generally the verbosity is usually opt-in, I feel that Corwin and Ash's direction (and the ticket that I believe Jane will work on after this) is explicitly to be _more loud_ and not less loud.
https://github.com/near/devx/issues/212

janedegtiareva: Are we fine with the options disappearing from everywhere though? It's a change from existing behaviour, and some folks don't like change

janedegtiareva: done

janedegtiareva:
https://github.com/near/near-shell/pull/440

PR #443, opened 2020-07-02
refactor: remove obsolete --vanilla flag

chadoh:
near/create-near-app#387 removed the --vanilla flag and made vanilla JS the default; we can remove this flag here


chadoh:
marking draft until near/create-near-app#387 is merged & deployed

mikedotexe:
Thanks, Chad, this has been fixed somewhere else. Closing

PR #444, opened 2020-07-03
fix: update test with create-near-app assumptions

chadoh:
the structure of create-near-app changed in v1 and we need to change this test to match

the commit that changed it is here: https://github.com/near/create-near-app/commit/7ab9ae34479dd098edf609b83ac78484c7b564fa

### DIFF
--- a/test/test_account_operations.sh
+++ b/test/test_account_operations.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 rm  -rf tmp-project
-yarn create near-app --vanilla tmp-project
+yarn create near-app tmp-project
 cd tmp-project
 timestamp=$(date +%s)
 testaccount=testaccount$timestamp.test.near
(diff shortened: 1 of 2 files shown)


chadoh:
@janedegtiareva I think this will fix the test (Travis didn't complete yet when I wrote this)

However, I think it would probably make sense to add `create-near-app` to this project's `devDependencies` and use the local version, rather than using the latest version with `yarn create`. Do you think it makes sense to do that before we merge this one?

janedegtiareva:
I think that the commit that fixes the --vanilla issue is needed too. I think that we also need to file a docs cleanup issue to remove --vanilla from all the things

chadoh:
Vanilla is not default until https://github.com/near/create-near-app/pull/387 is merged and published 

Note that docs changes are already in draft waiting on that one

It needs a review! Please help if you can

I still think having create-near-app unversioned in these tests is a bad idea. Let’s add it to devDependencies so future interface changes don’t break near-shell until dependabot bumps the version

Issue #410, opened 2020-06-08
Output explorer link for any successful transaction


eriktrautman:
For usability, it'd be very helpful if any confirmed transaction from the CLI outputs the tx ID with a helpful link to the explorer where it can be viewed.

evgenykuzyakov:
Moved to near-shell

evgenykuzyakov:
I'd also suggest to output explorer link for unsuccessful transactions. Because if the RPC node times out, it still possible that the TX was accepted.

janedegtiareva:
Printing txn id/ explorer link for failed txns requires near-api-js changes and makes this effort at least 5 story points.

janedegtiareva:
Let's make a different issue for printing it for errors, since it's more complex and requires near-api-js changes.

janedegtiareva:
https://github.com/near/near-shell/pull/463
https://github.com/near/near-api-js/pull/362

PR #448, opened 2020-07-09
Add a util for explorer functionality + unit tests

janedegtiareva:
near send will now print a link to explorer url at the end of the output, if there is a known explorer url and also if transaction id is known. 

This is needed for https://github.com/near/near-shell/issues/410

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "./test/index.sh",
+    "test": "jest && ./test/index.sh",
     "lint": "eslint .",
     "fix": "eslint . --fix"
   },
@@ -26,7 +26,7 @@
   },
   "devDependencies": {
     "eslint": "^7.0.0",
-    "jest": "^26.0.1",
+    "jest": "^26.1.0",
     "strip-ansi": "^6.0.0",
     "strip-ansi-cli": "^2.0.0"
   },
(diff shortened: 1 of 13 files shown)


mikedotexe:
This is great! I think the full log is more verbose than we want to be. I think the transaction hash is all that we really want if I'm understanding this ticket correctly.

Screenshot 2020-07-09 14 29 44

Screenshot 2020-07-09 14 27 30

Link worked great to me :)

package.json
```diff
@@ -7,7 +7,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "./test/index.sh",
+    "test": "jest && ./test/index.sh",
```
mikedotexe, line 10: this is causing tests to fail

mikedotexe:
Looks like there's linting issues in three files, too

package.json
janedegtiareva: It's a different issue (1 unfixed reference after refactoring). Addressed.

janedegtiareva:
> Looks like there's linting issues in three files, too

I already fixed linter issues

janedegtiareva:
re full response too verbose: I agree, and it also is not very user friendly. I was going to address that in a follow up PR so that the actual explorer work lands with a smaller surface area.

mikedotexe:
> re full response too verbose: I agree, and it also is not very user friendly. I was going to address that in a follow up PR so that the actual explorer work lands with a smaller surface area.

I think it's got everything we need, just call `getTxnId` so it says
Transaction ID: 234kljsdklfjk3l43
View this transaction:
https://explorer.…

PR #451, opened 2020-07-10
Add transaction result feedback for deploy and dev-deploy

janedegtiareva:


index.js
```diff
@@ -37,7 +37,14 @@ exports.deploy = async function (options) {
     const near = await connect(options);
     const contractData = [...fs.readFileSync(options.wasmFile)];
     const account = await near.account(options.accountId);
-    await account.deployContract(contractData);
+    const result = await account.deployContract(contractData);
+    if (options.verbose) {
```
vgrichina, line 41: any reason not to have this logic inside of `prettyPrintResponse`? So that code here can be something like:

```
prettyPrintResponse(options, await account.deployContract(contractData));
console.log(`Done deploying to ${options.accountId}`);
```

janedegtiareva: I like the refactoring suggestions, but we need to make a new fn instead of changing existing one. Existing one will still be needed for e.g. view calls

PR #469, opened 2020-07-23
Remove devnet from config since it's no longer running.

janedegtiareva:
Testnet can be used as a replacement for devnet.


janedegtiareva:
https://github.com/near/devx/issues/244

PR #471, opened 2020-07-24
Fix missing transaction id on call methods

janedegtiareva:
Tested: `node bin\near call stakingpool.youlaiwuqu.betanet ping '{}' --accountId jane.betanet`

```
Scheduling a call: stakingpool.youlaiwuqu.betanet.ping({})
Transaction Id Av6RZjt6PvXm9mcJL4jzcA18ojLCPY2yfoJZ9G6R7GPe
To see the transaction in the transaction explorer, please open this url in your browser
https://explorer.betanet.near.org/transactions/Av6RZjt6PvXm9mcJL4jzcA18ojLCPY2yfoJZ9G6R7GPe
```

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -33,4 +33,5 @@ async function scheduleFunctionCall(options) {
         utils.format.parseNearAmount(options.amount));
     const result = providers.getTransactionLastResult(functionCallResponse);
     inspectResponse.prettyPrintResponse(functionCallResponse, options);
+    console.log(inspectResponse.formatResponse(result));
 }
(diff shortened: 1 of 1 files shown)


commands/call.js
```diff
@@ -32,5 +32,6 @@ async function scheduleFunctionCall(options) {
         options.gas,
         utils.format.parseNearAmount(options.amount));
     const result = providers.getTransactionLastResult(functionCallResponse);
-    inspectResponse.prettyPrintResponse(result, options);
+    inspectResponse.prettyPrintResponse(functionCallResponse, options);
+    console.log(inspectResponse.formatResponse(result));
```
vgrichina, line 36: shouldn't this happen inside of `inspectResponse.prettyPrintResponse`? Seems like applies to every call that involves sending transaction.

janedegtiareva: Not all the code sending txns was calling getTransactionLastResult and printing it. Do we want to change it?

vgrichina: got it, seems like we don't want to change.

PR #481, opened 2020-08-03
Deprecate near-shell package, use near-cli package.

mikedotexe:
See https://github.com/near/devx/issues/237
and
https://github.com/near/devx/issues/236 (duplicate)

Deprecation warning up here:
https://www.npmjs.com/package/near-shell

New package here:
https://www.npmjs.com/package/near-cli

I had to use `1.0.1` instead of `1.0.0` as there was already an existing version there.
Combing through the various repos now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
-# NEAR Shell command line interface
+# NEAR CLI (command line interface)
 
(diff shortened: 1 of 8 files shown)


mikedotexe:
It appears that Travis has gotten confused as the repo was renamed after this PR was open. Closing it and will create a new PR on the same branch.

PR #496, opened 2020-08-12
create-account: console log the key file location

### LINKED ISSUE #490: create-account should print the path where it stored the access key
**Describe the feature**

I am as a regular user of near-shell had a problem finding the access key after I created it. It would be very helpful if near-shell tells me where it saved the access key.

**To Reproduce**

1. `env NEAR_ENV=local near --nodeUrl http://localhost:3030 --keyPath ~/.near/localnet/validator_key.json create-account new-account7.test.near --masterAccount test.near` (any other `create-account` is affected just the same, I just want to share the exact command I run)
2. See that there is no indication about the saved access keys at all, and there is no follow-up action suggested to me (`send` and `deploy` command examples would be helpful)

**Expected behavior**

The messages of CLI tool should navigate users through DevX.

mikedotexe:
Fixes #490 

Simple console log of the location of the file. In `near-api-js` turns out `getKeyFilePath` is private, so had to figure it out a different way.

### DIFF
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -104,6 +104,8 @@ async function createAccount(options) {
     let near = await connect(options);
     let keyPair;
     let publicKey;
+    let keyRootPath;
+    let keyFilePath;
     if (options.publicKey) {
         publicKey = options.publicKey;
     } else {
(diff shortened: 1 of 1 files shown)


commands/create-account.js
```diff
@@ -104,13 +104,21 @@ async function createAccount(options) {
...
         keyPair = await KeyPair.fromRandom('ed25519');
         publicKey = keyPair.getPublicKey();
     }
     if (keyPair) {
+        // keyFilePath = near.connection.signer.keyStore.getKeyFilePath(options.networkId, options.accountId);
+
+        if (near.connection.signer.keyStore.keyStores.length) {
+            keyRootPath = near.connection.signer.keyStore.keyStores[0].keyDir;
+        }
+        keyFilePath = `${keyRootPath}/${options.networkId}/${options.accountId}.json`;
```
frol, line 121: This seems to may lead to `undefined/testnet/...`

commands/create-account.js
```diff
@@ -127,6 +135,7 @@ async function createAccount(options) {
         }
     }
     // Create account
+    console.log(`Saving key to '${keyFilePath}'`);
```
frol, line 138: `keyFilePath` can be undefined here...

PR #502, opened 2020-08-18
check for account existence first

mikedotexe:
Quick fix for issue described here:

> earlier today I was using near cli to create a subaccount of the
> masterAccount in betanet. I came across of what seems to be as a
> critical error:
> 1) I accidentally ran "near create-account .." twice. And naturally got the error
> 'Sorry, account \'bugged.validator_italia.betanet\' already exists.'
> Nothing strange so far.
> 2)
>  However, after this error I was locked out of my subaccount and
> couldn't get my fonds out. I even tried copying the .json files with the
>  keys to a different server (which is still using near shell) and still
> couldn't sign any transactions with my subaccount. The error message I
> get is now
> 'Can not sign transactions for account bugged.validator_italia.betanet, no matching key pair found in Signer.',
> 3)
>  I tried this workflow with 3 different subaccounts and every time I was
>  locked out of the account. I attach a picture of the workflow to this
> email

### DIFF
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -112,18 +112,6 @@ async function createAccount(options) {
         keyPair = await KeyPair.fromRandom('ed25519');
         publicKey = keyPair.getPublicKey();
     }
-    if (keyPair) {
-        // keyFilePath = near.connection.signer.keyStore.getKeyFilePath(options.networkId, options.accountId);
-
-        if (near.connection.signer.keyStore.keyStores.length) {
-            keyRootPath = near.connection.signer.keyStore.keyStores[0].keyDir;
-        }
-        keyFilePath = `${keyRootPath}/${options.networkId}/${options.accountId}.json`;
-        await near.connection.signer.keyStore.setKey(options.networkId, options.accountId, keyPair);
-        await eventtracking.track(eventtracking.EVENT_ID_CREATE_ACCOUNT_END, { success: true, new_keypair: true }, options);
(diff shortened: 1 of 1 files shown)


48cfu:
So this was fixed at near cli level? Does this mean that if I use anyway near cli version 1.0.2, as when I originally discovered the bug, the error will present itself again?
Is it possible to fix it at a "blockchain level" in such a way that the error can't be reproduced regardless of the version of near cli?

mikedotexe:
> So this was fixed at near cli level? Does this mean that if I use anyway near cli version 1.0.2, as when I originally discovered the bug, the error will present itself again?
> Is it possible to fix it at a "blockchain level" in such a way that the error can't be reproduced regardless of the version of near cli?

Thank you for reporting this, we really appreciate it. 🙌🏼 
This was an error on the CLI level this time, and I believe it was my fault. Ironically, this is in the very place where I originally added code to make sure this exact problem wouldn't happen. Then mistakenly moved a block of code up. What's happening here was this:
Creating the account the second time overwrites the key pair file with the new private/public key, then called the blockchain to create the account, only to get an error that it already exists. This is fundamentally a NodeJS issue dealing with the unencrypted keystore saved to `~/.near-credentials`
This PR makes sure we check first for the existence of the account before ever creating the new keypair and saving them. It was a simple order of operations problem that I overlooked. Very thankful you pointed this out. If you have further questions, happy to answer

PR #508, opened 2020-08-23
fix: add test cleanup, fixes #507

### LINKED ISSUE #507: a cancelled/failed "yarn test" fails to remove ./tmp-project
***Describe the bug***
 If you execute `yarn test` and the test is cancelled or fails, all future `yarn test` will fail at the start

***To Reproduce***
 1\. run `export NODE_ENV=betanet && yarn test`
 2\. ***OR*** `yarn test` and cancel with ^C in the middle of a test, after `+yarn create near-app tmp-project`
 3\. run `yarn test` 
 4\. test fails at the start, because temporary `./tmp-project` directory was not deleted

***Expected behavior***
 `yarn test` running again

***Desktop:***

*   OS: GNU/Linux
*   Version 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64

***Additional context***
 The problem seems to be that a canceled test leaves a temporary `./tmp-project` directory
 It should be removed before the test begins

luciotato:
Fixes #507 by adding a cleanup call before test start

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "jest && ./test/index.sh",
+    "test": "./test/cleanup.sh && jest && ./test/index.sh",
     "lint": "eslint .",
     "fix": "eslint . --fix"
   },
--- a/test/index.sh
+++ b/test/index.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 export NODE_ENV=${NODE_ENV:-test}
+if [ "$NODE_ENV" == "betanet" -o "$NODE_ENV" == "development" -o "$NODE_ENV" == "testnet" ]; then
+    echo NODE_ENV=$NODE_ENV
+    echo -e "\e[31mFAIL\e[0m tests won't run if NODE_ENV=betanet|development|testnet"
+    echo please export NODE_ENV=test before testing
+    exit 1
+fi
 OVERALL_RESULT=0
 mkdir ~/.near-config
 echo '{"trackingEnabled":false}' >  ~/.near-config/settings.json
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "jest && ./test/index.sh",
+    "test": "./test/cleanup.sh && jest && ./test/index.sh",
```
vgrichina, line 10: @luciotato I think it's best to call `./test/cleanup.sh` from `./test/index.sh`

test/index.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 export NODE_ENV=${NODE_ENV:-test}
+if [ "$NODE_ENV" == "betanet" -o "$NODE_ENV" == "development" -o "$NODE_ENV" == "testnet" ]; then
```
vgrichina, line 3: that is not intended, tests should generally be possible to run on other environments, but might need to be adjusted

e.g. we run on `ci` and `ci-betanet` for now https://github.com/near/near-cli/blob/b3f5bde78451901d1e6cc8392e2ef2fc0bab23b9/.travis.yml#L6 but ideally we should be able to run as is on `testnet` / `betanet`

vgrichina (approved):
Looks good overall, there are some minor comments

package.json
luciotato: I think the `tmp-project` dir needs to be removed before calling jest
also if we leave it before jest, `./test/cleanup.sh` can be used to group all the test-cleanup tasks, not just fixing this problem.

test/index.sh
luciotato: I agree with you, but right now tests fail if NODE_ENV=betanet|development|testnet, and it's hard to see **why** tests are failing. It took me some time to see that the problem was my NODE_ENV. I added this to avoid the same pitfall for others.
I agree with you that if tests are adjusted, this warning should also be updated or removed.

luciotato:
@vgrichina can you merge this? I've no write access

mikedotexe:
Sorry this got stale, Lucio. I see the remote branch is removed. My team is taking over tooling and we'll be more responsive.

PR #529, opened 2020-09-07
Implement add-key option

### LINKED ISSUE #513: near-cli should support add key
Currently near-cli only supports deleting access keys, but doesn't support adding access keys.

ilblackdragon:
Fixes #513 

Usage:

Adding full access key to `illia` account:
```
near add-key CaSbwUJy2WEKKHHYnd6F7ZXUVUJ3AAzLBbDmvqML3jBg --account-id illia
```

Adding function access key to `illia` account for contract `test`, method name `deploy`, allowance: `1` $NEAR:
```
./bin/near add-key CaSbwUJy2WEKKHHYnd6F7ZXUVUJ3AAzLBbDmvqML3jBg --account-id illia --contract-id test --method-names deploy --allowance 1
```

PS. Where are we suppose to update documentation for this? @mikedotexe @potatodepaulo @amgando

### DIFF
--- a/commands/add-key.js
+++ b/commands/add-key.js
@@ -1,0 +1,44 @@
+const BN = require('bn.js');
+const exitOnError = require('../utils/exit-on-error');
+const connect = require('../utils/connect');
+const inspectResponse = require('../utils/inspect-response');
+const { utils } = require('near-api-js');
+
+module.exports = {
+    command: 'add-key [access-key]',
+    desc: 'Add an access key to given account',
+    builder: (yargs) => yargs
+        .option('access-key', {
+            desc: 'Public key to add (base58 encoded)',
+            type: 'string',
+            required: true,
+        })
+        .option('contract-id', {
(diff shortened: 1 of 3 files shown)


commands/add-key.js
```diff
@@ -0,0 +1,43 @@
...
+        .option('allowance', {
+            desc: 'Allowance in $NEAR for the key (default 0)',
+            type: 'string',
+            required: false,
+        }),
+    handler: exitOnError(addAccessKey)
+};
+
+async function addAccessKey(options) {
+    if (!options.accountId) {
```
vgrichina, line 34: @ilblackdragon shouldn't this throw error?

vgrichina: I'll change to throw error unless there are any objections.

vgrichina: or maybe just change to `add-key <account-id> <access-key>` as both seem to be require arguments

ilblackdragon: I just copied this from delete-key -> so prob should change there too.

Issue #523, opened 2020-09-02
generate-key --useLedgerKey should not take account Id


ilblackdragon:
Currently account Id is not used when calling `generate-key <accountId> --useLedgerKey`. 
 If should not require this option if `--useLedgerKey` is present.

PR #531, opened 2020-09-07
Enchancing Generate key

### LINKED ISSUE #524: generate-key should support an implicit account creation
We are adding implicit account creation, where account_id = `hex(public_key)`.
Hence, there should be a way to create such account via `generate-key`.

This should create new key pair, and based on it, create the file in credentials with account id that matches the encoded key.

Additionally, people may want to use custom mnemonic phrase for their keys. E.g. there should be a way to `near generate-key --mnemonic="tree forest ... bark"`. This way instead of generating new key pair, it will use this mnemonic (and possibly additional HD path) to get public key.

ilblackdragon:
Fixes #524 #523 

Supports next usages:
```
$ near generate-key --useLedgerKey
$ near generate-key --seed-phrase="seed phrase"
$ near generate-key account_id --seed-phrase="seed phrase"
```


commands/generate-key.js
```diff
@@ -24,8 +44,17 @@ module.exports = {
             return;
         }
 
-        const keyPair = KeyPair.fromRandom('ed25519');
-        await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
-        console.log(`Generated key pair with ${keyPair.publicKey} public key`);
+        let publicKey, accountId;
+        if (argv.seedPhrase) {
+            const result = parseSeedPhrase(argv.seedPhrase, argv.seedPath);
+            publicKey = result.publicKey;
```
evgenykuzyakov, line 50: Is this a string or `PublicKey`?

ilblackdragon: it's a string.

ilblackdragon: https://github.com/near/near-seed-phrase/blob/master/index.js#L18

commands/add-key.js
```diff
@@ -0,0 +1,43 @@
+const exitOnError = require('../utils/exit-on-error');
+const connect = require('../utils/connect');
+const inspectResponse = require('../utils/inspect-response');
+const { utils } = require('near-api-js');
+
+module.exports = {
```
vgrichina, line 6: looks like spillover from other PR

commands/generate-key.js
vgrichina: seems like makes sense to handle this in middleware, similar to Ledger:
https://github.com/near/near-cli/blob/c6870c5f30799cb8cdcdc55046edba792d37daa6/middleware/ledger.js#L46

PR #532, opened 2020-09-07
added informative error when account is not found

rozum-dev:


index.js
```diff
@@ -169,14 +169,35 @@ exports.login = async function (options) {
...
+    try {
+        let near = await connect(options);
+        let account = await near.account(options.accountId);
+        let state = await account.state();
+        if (state && state.amount) {
+            state['formattedAmount'] = utils.format.formatNearAmount(state.amount);
+        }
+        console.log(`Account ${options.accountId}`);
+        console.log(inspectResponse.formatResponse(state));
+    } catch(e) {
```
vgrichina, line 181: best to move this error handling into https://github.com/near/near-cli/blob/master/utils/exit-on-error.js so  that  it applies to all commands

index.js
```diff
@@ -169,14 +169,35 @@ exports.login = async function (options) {
...
+        console.log(chalk`\n{bold.red Account {bold.white ${options.accountId}} is not found in {bold.white ${config.helperAccount}} network.\n}`);
```
vgrichina, line 182: need to check whether error is actually about account being not found (there can be other errors, like e.g. no Internet connection)

index.js
```diff
@@ -169,14 +169,35 @@ exports.login = async function (options) {
...
+
+        const prefix = String(options.accountId).match(/[^\.]*$/gi)[0];
```
vgrichina, line 184: nit: seems to be suffix, not prefix

rozum-dev: I just called it as it's been described in issue description, will rename np

index.js
rozum-dev: The same as I described above. 
Is there some doc on server errors response so I don't have to guess what happened.

index.js
rozum-dev: You are absolutely right that it would be better, but I can see potential problem here.
If I transfer the logic to this file and the user, for example runs the "sendMoney" command (or any other - where there is no accountId, etc.), and the server returns a weird error (for example, Currently error is "[-32000] Server error: ... "with huge stack trace.) then we get a weird wrong error case. I don't know witch cases can potentially cause "[-32000] Server error:"
How would you suggest manage those cases?

vgrichina: As far as I know the error for account being not found is:

```
TypedError: [-32000] Server error: account bazinga does not exist while viewing
```

i.e. you can check for "does not exist while viewing" substring.

utils/inspect-response.js
```diff
@@ -1,6 +1,31 @@
...
+        case 'testnet': 
+            console.log(chalk`{bold.white Use export NEAR_ENV=testnet to use TestNet accounts. \n}`);
+            break;
+        case 'betanet': 
+            console.log(chalk`{bold.white Use export NEAR_ENV=betanet to use BetaNet accounts. \n}`);
+            break;
+    }
+
+    return true;
+}
```
volovyk-s, line 27: another nit-pick: missing semicolon. Please, check output from Travis CI, there is a banch of other minor issues:
`12:55  error  Unnecessary escape character: \.               no-useless-escape
  15:1   error  Expected indentation of 4 spaces but found 8   indent
  16:1   error  Expected indentation of 8 spaces but found 12  indent
  17:1   error  Expected indentation of 8 spaces but found 12  indent
  18:1   error  Expected indentation of 4 spaces but found 8   indent
  19:1   error  Expected indentation of 8 spaces but found 12  indent
  20:1   error  Expected indentation of 8 spaces but found 12  indent
  21:1   error  Expected indentation of 4 spaces but found 8   indent
  22:1   error  Expected indentation of 8 spaces but found 12  indent
  23:1   error  Expected indentation of 8 spaces but found 12  indent
  27:2   error  Missing semicolon                              semi`

volovyk-s (changes requested):
Please, fix these minor issues before we will approve this PR. We should manage errors in a more systematic way, but probably it should be a part of a separate PR.

utils/inspect-response.js
```diff
@@ -1,6 +1,32 @@
...
+const chalk = require('chalk');  // colorize output
 const util = require('util');
+
+
+const checkExistAccount = (error, options) => {
+    if(!String(error).includes('does not exist while viewing')) return false;
+
+    console.log(chalk`\n{bold.red Account {bold.white ${options.accountId}} is not found in {bold.white ${config.helperAccount}} network.\n}`);
+    
+    const re = new RegExp("[^\.]*$", "gi")
```
volovyk-s, line 12: @rozum-dev There is still several issues in this line ;)

`12:27  error  Strings must use singlequote quotes`
`12:30  error  Unnecessary escape character: \. no-useless-escape`
`12:38  error  Strings must use singlequote quotes`
`12:43  error  Missing semicolon semi`

Consider using `yarn run lint` to check for such issues and `yarn run fix` to fix them (most of them).

volovyk-s (changes requested):

utils/inspect-response.js
```diff
@@ -1,6 +1,32 @@
...
+const chalk = require('chalk');  // colorize output
 const util = require('util');
+
+
+const checkExistAccount = (error, options) => {
+    if(!String(error).includes('does not exist while viewing')) return false;
+
+    console.log(chalk`\n{bold.red Account {bold.white ${options.accountId}} is not found in {bold.white ${config.helperAccount}} network.\n}`);
+    
+    const re = new RegExp('[^\.]*$', 'gi');
```
volovyk-s, line 12: I'm not that picky, but Travis CI is. Only one issue left!
Please, change to `[^\.]*$` to ` [^.]*$`. That escape character is redundant.

P.S. You can check CI status in the `Checks` tab of the PR. Usually it's ready in 10-15 minutes after the commit.

volovyk-s (changes requested):

volovyk-s:
@rozum-dev Thank you for your contribution!

PR #535, opened 2020-09-09
Feature: Add 2FA Support to CLI

mattlockyer:
WIP can send transfers using 2fa account with --use2fa

TODO
1. Update convertActions helper to support deployContract
2. Test more CLI methods e.g. deploy


mattlockyer:
WIP UX for send command
image

ilblackdragon:
Is there a way to not need --2fa flag? Ideally we should figure out it from account itself.

Multisig is easily detectable. Not sure if we want to expose CH to tell if this account is managed by it or not - looks a bit like of privacy/attack issue.

Also Account object should work not just specifically to 2fa but for any multisig account. This probably can even live in near-api-js eventually.

And 2FA should be just extra wrapper over Account that if it's 2fa account - after request was created it will wait for confirmation and query for result after tx confirmed.

mattlockyer:
> Is there a way to not need --2fa flag? Ideally we should figure out it from account itself.

Yes we can check hashes or use CH

> Multisig is easily detectable. Not sure if we want to expose CH to tell if this account is managed by it or not - looks a bit like of privacy/attack issue. Right now it's exposed just by the fact that the key is the same, but we would ideally generate new public key per user going forward.

Each time you login with wallet you get a 2FA key for the CLI. This key is multisig key only. You need to provide signature to access any CH methods that expose recovery / 2FA details so I think it's secure?

> Also Account object should work not just specifically to 2fa but for any multisig account. This probably can even live in near-api-js eventually.

Agree with this point here. Will take a bit of time to think about proper abstractions. Appreciate any input here: @vgrichina 

> And 2FA should be just extra wrapper over Account that if it's 2fa account - after request was created it will wait for confirmation and query for result after tx confirmed.

Sure. Should we continue with `extends Account` inheritance pattern to wrap functionality of Account for special contract accounts?

dentesting:
I think by looking at the storage used (2FA vs none-2FA) and keys, one can easily identify its a 2FA account: 
Screenshot 2020-09-25 at 11 46 53

mattlockyer:
@dentesting thanks for joining this discussion.

We decided to move the mutlisig support to `near-api-js`:
https://github.com/near/near-api-js/pull/403

There is already a wallet PR that uses this:
https://github.com/near/near-wallet/pull/914

This will allow us to create a simple CLI PR that detects if the account has a multisig contract (using access keys and method names or contract hashes as it's done in the wallet).

ETA for CLI PR - next week

dentesting:
> @dentesting thanks for joining this discussion.
> ETA for CLI PR - next week
Excellent, I will be happy to test.

mattlockyer:
Works with:
https://github.com/near/near-api-js/pull/414

package.json
```diff
@@ -45,6 +45,7 @@
     "near-api-js": "^0.30.0",
     "near-seed-phrase": "^0.1.0",
     "open": "^7.0.1",
+    "prompts": "^2.3.2",
```
vgrichina, line 48: we are using `readline` in other places when prompting user. I think we either need to update everything to prompts or continue using `readline`

utils/connect.js
```diff
@@ -1,6 +1,33 @@
...
-    // TODO: Avoid need to wrap in deps
-    return await nearConnect({ ...options, deps: { keyStore }});
+    const near = await nearConnect({ ...options, deps: { keyStore }});
+    near.account = async (accountId) => {
+        const account = new Account(near.connection, accountId);
+        await account.state();
+        const accessKeys = await account.getAccessKeys();
+        const hasFAK = accessKeys.find((k) => k.access_key.permission && k.access_key.permission === 'FullAccess');
+        if (hasFAK) {
+            return account;
```
vgrichina, line 18: shouldn't we check whether full access key actually exists locally? Otherwise returned `account` going to be unusable.

vgrichina: e.g. this will happen for all seed phrase + 2FA accounts

utils/connect.js
```diff
@@ -1,6 +1,33 @@
...
+        MULTISIG_CHANGE_METHODS
+    }
+} = require('near-api-js');
+const { AccountMultisig } = require('./2fa');
 
 module.exports = async function connect({ keyStore, ...options }) {
-    // TODO: Avoid need to wrap in deps
-    return await nearConnect({ ...options, deps: { keyStore }});
+    const near = await nearConnect({ ...options, deps: { keyStore }});
+    near.account = async (accountId) => {
```
vgrichina, line 12: seems like this logic should be in `near-api-js`

utils/connect.js
```diff
@@ -1,6 +1,33 @@
...
+        const use2fa = accessKeys.find((k) => 
+            k.access_key.permission && 
+            k.access_key.permission.FunctionCall &&
+            k.access_key.permission.FunctionCall.method_names.some((mn) => MULTISIG_CHANGE_METHODS.includes(mn))
+        );
+        if (use2fa) {
+            return new AccountMultisig(near.connection, accountId);
+        }
+
+        throw Error('No account matching', accountId);
```
vgrichina, line 30: This error message would make me think that account doesn't exist at all, which isn't true. Also `accountId` is basically passed as `fileName` parameter of `Error` constructor, which is probably unintentional. 

I think we should do something more like:

```
throw new Error(`Account ${accountId} doesn't have either full access keys or multisig`);
```

however I think even then we also need to support case where there is local limited access key as you can use it with `near call` or `near repl`

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+            await this.sendRequestCode();
+            // say something and wait for input
+            const response = await prompts({
+                type: 'text',
+                name: 'securityCode',
+                message: `Enter security code received via ${ method.kind.split('2fa-')[1] }`
+            });
+            console.log('verifying security code ...');
+            const res = await this.verifyRequestCode(response.securityCode);
+            if (res.success) {
```
vgrichina, line 37: nit: this kind of conditions look better with early return, e.g.

```js
if (!res.success) {
   throw new Error('Invalid security code');
}
if (typeof res.res !== 'string' && res.res === false) {
     throw new Error('Request failed with error:', res);
}
console.log('Request confirmed with return value:', res.res);
```

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+                if (typeof res.res === 'string' || res.res !== false) {
```
vgrichina, line 38: nit: why not `result` to avoid confusion on what is `res`?

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+                    console.log('Request confirmed with return value:', res.res);
```
vgrichina, line 39: what is a return value of the request? If this is whatever transaction returned – shouldn't it just be returned from this method?

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+                } else {
+                    throw new Error('Request failed with error:', res);
```
vgrichina, line 41: why not throw error from `verifyRequestCode`?

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+                }
+            } else {
+                throw new Error('Invalid security code');
```
vgrichina, line 44: why not throw error from `verifyRequestCode`?

also seems like need to allow user to retry?

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+            }
+        } catch (e) {
+            console.log(e);
```
vgrichina, line 47: not sure why not just let error pass through back to caller (i.e. why this `catch` is needed)

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
...
+    multisig: {
+        AccountMultisig: AccountMultisigApi,
+    },
+} = require('near-api-js');
+
+class AccountMultisig extends AccountMultisigApi {
+    constructor(connection, accountId) {
+        super(connection, accountId);
+    }
+    // override for custom UX
```
vgrichina, line 13: how does it work when we don't override? Does it make sense to allow to provide some callbacks for input when creating `Account2FA`?

utils/2fa.js
```diff
@@ -0,0 +1,55 @@
+
+const prompts = require('prompts');
+const {
+    multisig: {
+        AccountMultisig: AccountMultisigApi,
+    },
+} = require('near-api-js');
+
+class AccountMultisig extends AccountMultisigApi {
```
vgrichina, line 9: Seems like right now it is specific to 2FA, so makes sense to have `Account2FA`.

package.json
mattlockyer: Using readline and chalk now

utils/connect.js
mattlockyer: checking against
```
const localKey = (await near.connection.signer.getPublicKey(account.accountId, near.connection.networkId)).toString();
```

utils/2fa.js
mattlockyer: It will try to signAndSendTransaction without sending code from helper. You must override it and call helper methods, unless you have another way to call confirm on user's account.

package.json
vgrichina: change is still here though

utils/2fa.js
vgrichina: seems like `Account2FA ` should be able to pass callbacks instead, like:

```
new Account2FA(connection, accountId, {
     getSecurityCode: async function() {
         //  Loop to query user input in console goes here
     }
});
```

package.json
mattlockyer: I think it was cached? It's not in `package.json` anymore

vgrichina (approved):
this looks good, just make sure to address comments in https://github.com/near/near-api-js/pull/428#pullrequestreview-505280342

PR #557, opened 2020-09-30
build person profile

icerove:
add people.set function to the eventtracking.js to build user profile

### DIFF
--- a/utils/eventtracking.js
+++ b/utils/eventtracking.js
@@ -1,15 +1,15 @@
-const MIXPANEL_TOKEN = 'e98aa9d6d259d9d78f20cb05cb54f5cb';
+const MIXPANEL_TOKEN = "e98aa9d6d259d9d78f20cb05cb54f5cb";
 
-const chalk = require('chalk');  // colorize output
-const crypto = require('crypto');
-const mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
-const near_cli_version = require('../package.json').version;
-const readline = require('readline');
-const settings = require('./settings');
-const uuid = require('uuid');
+const chalk = require("chalk"); // colorize output
+const crypto = require("crypto");
+const mixpanel = require("mixpanel").init(MIXPANEL_TOKEN);
+const near_cli_version = require("../package.json").version;
+const readline = require("readline");
+const settings = require("./settings");
+const uuid = require("uuid");
 
-const TRACKING_ENABLED_KEY = 'trackingEnabled';
-const TRACKING_SESSION_ID_KEY = 'trackingSessionId';
+const TRACKING_ENABLED_KEY = "trackingEnabled";
+const TRACKING_SESSION_ID_KEY = "trackingSessionId";
 
 const isGitPod = () => {
     return !!process.env.GITPOD_WORKSPACE_URL;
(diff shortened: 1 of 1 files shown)


utils/eventtracking.js
```diff
@@ -40,41 +47,53 @@ const track = async (eventType, eventProperties, options) => {
...
                 rl.question(
                     chalk`We would like to collect data on near-cli usage to improve developer experience.` +
-                    chalk` We will never send private information. We only collect which commands are run via an anonymous identifier.` +
-                    chalk`{bold.yellow  Would you like to opt in (y/n)? }`,
+                        chalk` We will never send private information. We only collect which commands are run via an anonymous identifier.` +
+                        chalk`{bold.yellow  Would you like to opt in (y/n)? }`,
                     async (consentToEventTracking) => {
-                        if (consentToEventTracking == 'y' || consentToEventTracking == 'Y') {
+                        if (
+                            consentToEventTracking == 'y' ||
```
vgrichina, line 84: nit: just use `.toLowerCase()` on `consentToEventTracking`? Also why is this touched at all?

utils/eventtracking.js
```diff
@@ -90,11 +109,23 @@ const askForConsentIfNeeded = async (options) => {
     const shellSettings = settings.getShellSettings();
     // if the appropriate option is not in settings, ask now and save settings.
     if (!(TRACKING_ENABLED_KEY in shellSettings)) {
-        shellSettings[TRACKING_ENABLED_KEY] = shouldOptInByDefault() || await getEventTrackingConsent();
-        shellSettings[TRACKING_SESSION_ID_KEY] = shellSettings[TRACKING_ENABLED_KEY] ? uuid.v4() : undefined;
+        shellSettings[TRACKING_ENABLED_KEY] =
```
vgrichina, line 112: is this intentional change of formatting?

utils/eventtracking.js
```diff
@@ -90,11 +109,23 @@ const askForConsentIfNeeded = async (options) => {
...
+        shellSettings[TRACKING_SESSION_ID_KEY] = shellSettings[
+            TRACKING_ENABLED_KEY
+        ]
+            ? uuid.v4()
+            : undefined;
         settings.saveShellSettings(shellSettings);
         if (shellSettings[TRACKING_ENABLED_KEY]) {
-            track(module.exports.EVENT_ID_TRACKING_OPT_IN, {}, options);
+            await track(module.exports.EVENT_ID_TRACKING_OPT_IN, {}, options);
+            await mixpanel.people.set({
```
vgrichina, line 122: how does this work? If this is set globally, why is it also sent with `track`? Why call `mixpanel.people.set` after `track` and not before?

icerove: track function is for event tracking, and people.set function is setting user profile, it does not matter for the sequence in my mind. Is that matter? @vgrichina

utils/eventtracking.js
icerove: I think so, this is all after I command `yarn fix`

utils/eventtracking.js
icerove: I do not touch this, I think it is from eslint fixing

icerove:
Is this okay to merge ? @vgrichina

utils/eventtracking.js
```diff
@@ -40,41 +47,49 @@ const track = async (eventType, eventProperties, options) => {
...
-                    chalk`{bold.yellow  Would you like to opt in (y/n)? }`,
+                        chalk` We will never send private information. We only collect which commands are run via an anonymous identifier.` +
+                        chalk`{bold.yellow  Would you like to opt in (y/n)? }`,
                     async (consentToEventTracking) => {
-                        if (consentToEventTracking == 'y' || consentToEventTracking == 'Y') {
+                        if (consentToEventTracking.toLowerCase() == 'y') {
                             resolve(true);
-                        } else if (consentToEventTracking == 'n' || consentToEventTracking == 'N') {
+                        } else if (
+                            consentToEventTracking.toLowerCase() == 'n'
```
vgrichina, line 86: nit: formatting is a bit weird (no reason not to keep `if` and condition on same line)

vgrichina:
@icerove merged, thanks! Let me know if we need to cut a release

PR #581, opened 2020-10-20
generate function for asking id

icerove:
generate asking question askForId to ask user whether okay to expose their account id when typing account id with command. This is for mixpanel project of identity management 
change config 'default' with 'testnet'

### DIFF
--- a/utils/eventtracking.js
+++ b/utils/eventtracking.js
@@ -10,6 +10,7 @@ const uuid = require('uuid');
 
 const TRACKING_ENABLED_KEY = 'trackingEnabled';
 const TRACKING_SESSION_ID_KEY = 'trackingSessionId';
+const TRACKING_ID_KEY = 'trackingID';
 
 const isGitPod = () => {
     return !!process.env.GITPOD_WORKSPACE_URL;
@@ -37,6 +38,20 @@ const shouldTrack = (shellSettings) => {
     return (
         TRACKING_ENABLED_KEY in shellSettings &&
         shellSettings[TRACKING_ENABLED_KEY]
+    );
+};
+
+const shouldTrackID = (shellSettings) => {
+    return (
+        TRACKING_ID_KEY in shellSettings &&
+        shellSettings[TRACKING_ID_KEY]
+    );
+};
+
+const shouldNOTTrackID = (shellSettings) => {
+    return (
+        TRACKING_ID_KEY in shellSettings &&
+        !shellSettings[TRACKING_ID_KEY]
     );
 };
 
@@ -115,7 +130,7 @@ const getIdTrackingConsent = async () => {
         output: process.stdout,
     });
     try {
-        for (let attempts = 0; attempts < 10; attempts++) {
+        for (let attempts = 0; attempts < 3; attempts++) {
             const answer = await new Promise((resolve) => {
                 rl.question(
                     chalk`We would like to help with your development journey with NEAR.` +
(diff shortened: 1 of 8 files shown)


utils/eventtracking.js
```diff
@@ -109,6 +109,53 @@ const getEventTrackingConsent = async () => {
...
+    const rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+    });
+    try {
+        for (let attempts = 0; attempts < 10; attempts++) {
+            const answer = await new Promise((resolve) => {
+                rl.question(
+                    chalk`We would like to help with your development journey with NEAR.` +
+                        chalk` We will ask you to expose your account Id while using command. ` +
```
mikedotexe, line 122: Do you think we could replace "expose" with "share" here and the next line?

icerove: yes, definitely, good idea, thanks

mikedotexe:
It keeps asking me the same question after I say "no, thank you" by typing **n**
I was using this command:
`./bin/near state mike.testnet`

Screen Shot 2020-10-27 at 2 37 39 PM

icerove:
I will fix this, thanks

mikedotexe:
The tests seem to be failing as it's waiting for a response on this new prompt:

> ++../bin/near state testaccount1603834350.test.near -v
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
> Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
> The build has been terminated

icerove:
I see, I will fix the test, too. Thanks

mikedotexe:
Sorry, I hate to keep adding comments, but I'm trying all the different ways.
I think if the user says, "no thank you, please don't track me" we should not ask them again if they want to share their account ID.
For instance, if I do `near login` and it prompts me, and I say, "no thank you" and then do `near state` it will ask me if I want to share my account ID.
I think if the answer is no to the first question, the second question should not be asked, but that is not how it behaves now.

mikedotexe:
Unfortunately I cannot see Mixpanel because of a bug on their end, but I've filed a ticket. I think as long as the most recent comment is addressed and it's showing the person's accountId in Mixpanel, good to go!

icerove:
Thank you so much for testing and reviewing for me.
Actually, I consider the no situation here, and when the setting is like this: 
image

utils/eventtracking.js
```diff
@@ -109,6 +124,57 @@ const getEventTrackingConsent = async () => {
...
+        }
+        return false; // If they can't figure it out in this many attempts, just opt out
+    } finally {
+        rl.close();
+    }
+};
+
+const askForId = async (options) => {
+    const shellSettings = settings.getShellSettings();
+    if(shouldTrackID(shellSettings)){
```
icerove, line 163: this is when the answer is 'Y' or 'y'

utils/eventtracking.js
```diff
@@ -109,6 +124,57 @@ const getEventTrackingConsent = async () => {
...
+        const id = isGitPod() ? getGitPodUserHash() : shellSettings[TRACKING_SESSION_ID_KEY];
+        await Promise.all([
+            mixpanel.alias(options.accountId, id),
+            mixpanel.people.set_once({account_id: options.accountId})
+        ]);
+    }else if(shouldNOTTrackID(shellSettings)){
```
icerove, line 169: this is when answer is 'N' or 'n'

utils/eventtracking.js
```diff
@@ -109,6 +124,57 @@ const getEventTrackingConsent = async () => {
...
+        return;
+    }
+    else{
```
icerove, line 172: This is first time to ask for that

icerove:
Would you mind check where I point out again?

mikedotexe (approved):
awesome!

PR #585, opened 2020-10-26
Save account credentials with generate-key --seedPhrase option

kucharskim:
I was able to generate account.json credentials file from seed phrase,
received when I was initially setting up account via official NEAR web
wallet.

      env NODE_ENV=mainnet near generate-key --seedPhrase 'secret recovery phrase' account

I find it very convenient. This addresses https://github.com/near/near-cli/issues/570

### DIFF
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -35,6 +35,9 @@ module.exports = {
             argv.publicKey = keyPair.publicKey.toString();
             argv.accountId = argv.accountId || implicitAccountId(argv.publicKey);
             await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
+        } else if (argv.seedPhrase) {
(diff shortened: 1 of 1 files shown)


kucharskim:
If there is something more to test, let me know.

mikedotexe:
> If there is something more to test, let me know.

thank you so much! I'll try to get to this today

kucharskim:
> thank you so much! I'll try to get to this today

I think I would need to update also https://docs.near.org/docs/development/near-cli as that documentation would be incorrect with the behavior of the near-cli after this pull request is merged.

mikedotexe (approved):
great stuff, thank you!

mikedotexe:
> > thank you so much! I'll try to get to this today
> 
> I think I would need to update also https://docs.near.org/docs/development/near-cli as that documentation would be incorrect with the behavior of the near-cli after this pull request is merged.

cc @thisisjoshford

PR #594, opened 2020-11-07
Revert "generate function for asking id"

vgrichina:
Reverts near/near-cli#581


vgrichina:
Reverting as this breaks usage in batch scripts (as evident by the fact that tests needed to be updated).

There is existing `askForConsentIfNeeded` method where we should add text about account ID sharing consent. It is called only from commands which are known to be user-interactive and so doesn't interfere with batch usage or any other cases where user doesn't expect question (e.g. `near keys account.near | grep app.account.near`)

PR #603, opened 2020-11-17
feat: find wasm files built by cargo or asb

willemneal:
When deploying and "./out/main.wasm" option is used but the file is not present, this looks for `**/{target}/*.wasm` since both Rust and AssemblyScript projects produce binaries with that pattern.

`dev-deploy` looks for the debug target and `deploy` looks for release.

### DIFF
--- a/utils/find-wasm.js
+++ b/utils/find-wasm.js
@@ -1,28 +1,29 @@
-const glob = require('glob');
 const util = require('util');
+const glob = util.promisify(require('glob'));
+const fs = require('fs');
 
-let glob_promise = util.promisify(glob);
+
 
 async function findWasm(target) {
-  let res = await glob_promise(`**/${target}/*.wasm`, { 
-    ignore: "**/node_modules/**"
-  });
-  return res;
+    let res = await glob(`**/${target}/*.wasm`, { 
+        ignore: '**/node_modules/**'
+    });
+    return res;
 }
 
 async function findWasmFile(wasmFile, target) {
     if (!fs.existsSync(wasmFile)) {
-        if (wasmFile === "./out/main.wasm") {
-          let res = await findWasm(target);
-          if (res.length == 0) {
-            console.error(`Wasm file ${wasmFile} does not exist and none could be found for target ${target}.`);
-            process.exit(1);
-          }
-          if (res.length > 1) {
-            console.error(`Multiple Wasm files found ${res.join(", ")} please pass one with --wasmFile=file`);
-            process.exit(1);
-          }
-          return res[0];
+        if (wasmFile === './out/main.wasm') {
+            let res = await findWasm(target);
+            if (res.length == 0) {
+                console.error(`Wasm file ${wasmFile} does not exist and none could be found for target ${target}.`);
+                process.exit(1);
+            }
+            if (res.length > 1) {
+                console.error(`Multiple Wasm files found ${res.join(', ')} please pass one with --wasmFile=file`);
+                process.exit(1);
+            }
+            return res[0];
         } else {
             console.error(`Wasm file ${wasmFile} does not exist.`);
             process.exit(1);
--- a/utils/find-wasm.js
+++ b/utils/find-wasm.js
@@ -13,14 +13,14 @@ async function findWasm(target) {
 
 async function findWasmFile(wasmFile, target) {
     try {
-        return findWasmFileSafe(wasmFile, target);
+        return await findWasmFileUnsafe(wasmFile, target);
     } catch (err) {
         console.error(err.message);
         process.exit(1);
     }
 }
 
-async function findWasmFileSafe(wasmFile, target) {
+async function findWasmFileUnsafe(wasmFile, target) {
     if (fs.existsSync(wasmFile)) {
         return wasmFile;
     }
@@ -39,4 +39,4 @@ async function findWasmFileSafe(wasmFile, target) {
 }
 
 
-module.exports = {findWasmFile, findWasmFileSafe};
+module.exports = {findWasmFile, findWasmFileUnsafe};
--- a/utils/find-wasm.js
+++ b/utils/find-wasm.js
@@ -12,24 +12,31 @@ async function findWasm(target) {
 }
 
 async function findWasmFile(wasmFile, target) {
-    if (!fs.existsSync(wasmFile)) {
-        if (wasmFile === './out/main.wasm') {
-            let res = await findWasm(target);
-            if (res.length == 0) {
-                console.error(`Wasm file ${wasmFile} does not exist and none could be found for target ${target}.`);
-                process.exit(1);
-            }
-            if (res.length > 1) {
-                console.error(`Multiple Wasm files found ${res.join(', ')} please pass one with --wasmFile=file`);
-                process.exit(1);
-            }
-            return res[0];
-        } else {
-            console.error(`Wasm file ${wasmFile} does not exist.`);
-            process.exit(1);
-        }
-    }
-    return wasmFile;
+  try {
+    return fileWasmFileSafe(wasmFile, target);
+  } catch (err) {
+    console.error(err.message);
+    process.exit(1);
+  }
 }
 
-module.exports = findWasmFile;
+async function findWasmFileSafe(wasmFile, target) {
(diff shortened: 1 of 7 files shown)


utils/find-wasm.js
```diff
@@ -0,0 +1,35 @@
+const util = require('util');
+const glob = util.promisify(require('glob'));
+const fs = require('fs');
+
+
+
+async function findWasm(target) {
+    let res = await glob(`**/${target}/*.wasm`, {
```
chadoh, line 8: Does this pattern match `contract/rust/target/wasm32-unknown-uknown/debug/whatever.wasm`? If it matches the behavior I get in bash on a Mac, it does not.

utils/find-wasm.js
```diff
@@ -0,0 +1,35 @@
...
+    let res = await glob(`**/${target}/*.wasm`, { 
+        ignore: '**/node_modules/**'
+    });
+    return res;
+}
+
+async function findWasmFile(wasmFile, target) {
+    if (!fs.existsSync(wasmFile)) {
+        if (wasmFile === './out/main.wasm') {
+            let res = await findWasm(target);
```
chadoh, line 17: If `./out/main.wasm` is passed, we _ignore_ it and find a separate file instead? 

I don't think that's expectable behavior. I think we need to change/remove the default, rather than use it as a second-tier, undocumented flag.

To me, expectable behavior would be:

* if I provide an option, use it!
* if I don't:
  * use `./out/main.wasm` if it exists, whether for `dev-deploy` or `deploy`. This matches the current experience.
  * look for `contract/target/wasm32-unknown-unknown/${target}/*.wasm` and `contract/build/${target}/*.wasm`, if only one found, use it
    * if multiple wasm files found, throw error message listing these matched wasm files (we can update this later to use an interactive mode to select which you want to deploy)

utils/find-wasm.js
```diff
@@ -0,0 +1,35 @@
...
+            if (res.length == 0) {
+                console.error(`Wasm file ${wasmFile} does not exist and none could be found for target ${target}.`);
+                process.exit(1);
+            }
+            if (res.length > 1) {
+                console.error(`Multiple Wasm files found ${res.join(', ')} please pass one with --wasmFile=file`);
```
chadoh, line 23: This output seems less than ideal. What do each of the entries in `res` look like? Let's make sure they are relative paths, not absolute. And then, let's list each on a separate line, like this:

```js
                console.error(`Multiple Wasm files found: ${res.map(file => 
                  `\n  • ${file}`
                )}\n\nPlease pass one with --wasmFile=file`);
```

chadoh (changes requested):

utils/find-wasm.js
willemneal: Hmm I thought ** was recursive.  If you do `**/*/${target}/*.wasm` does it work?

utils/find-wasm.js
```diff
@@ -0,0 +1,35 @@
...
+                process.exit(1);
+            }
+            return res[0];
+        } else {
+            console.error(`Wasm file ${wasmFile} does not exist.`);
+            process.exit(1);
```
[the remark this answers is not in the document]
willemneal: This is the case that `wasmFile` passed doesn't exist and it isn't the default.

utils/find-wasm.js
willemneal: This does just that.  If the file exists, `wasmFile` is returned.  Otherwise if it was `./out/main.wasm` that doesn't exist we check the the other paths using target.

I'll rewrite this in a positive condition to make this clearer.

utils/find-wasm.js
willemneal: I just added tests that show this does work.

utils/find-wasm.js
```diff
@@ -0,0 +1,42 @@
...
+        console.error(err.message);
+        process.exit(1);
+    }
+}
+
+async function findWasmFileSafe(wasmFile, target) {
+    if (fs.existsSync(wasmFile)) {
+        return wasmFile;
+    }
+    if (wasmFile === './out/main.wasm') {
```
chadoh, line 27: I still think we should probably remove the default from the `deploy` and `dev-deploy` commands, and add some sort of concise description to the help output for these commands, explaining this behavior. But maybe it's worth getting a second opinion.

@mikedotexe what do you think? This change is going to make it possible to remove the symlinking-to-`out/main.wasm` behavior from `create-near-app`'s `compile` scripts, and just find the single `*.wasm` file in the `contract` folder. It also sets us up to add interactivity to near-cli later, so if you have a project with multiple contracts, it can list your Wasm files for you and ask which you'd like to deploy. 

Right now, the help docs for `near deploy --help` and `near dev-deploy --help` say:

```
--wasmFile, --wasm_file         Path to wasm file to deploy  [string] [default: "./out/main.wasm"]
```

Should we remove the default now? Should we make it say something like:

```
--wasmFile, --wasm_file         Path to wasm file to deploy. Unless specified, looks for "./out/main.wasm" or "**/release/*.wasm"  [string]
```

mikedotexe: I feel like leaving the default makes sense as we have a number of examples that'll work seamlessly. Feels like what you're suggesting might add one (small) extra step

utils/find-wasm.js
```diff
@@ -0,0 +1,42 @@
...
+        let res = await findWasm(target);
+        if (res.length == 0) {
+            throw new Error(`Wasm file ${wasmFile} does not exist and none could be found for target ${target}.`);
+        }
+        if (res.length > 1) {
+            throw new Error(`Multiple Wasm files found:\n\t• ${res.join('\n\t• ')}\n\nPlease pass one with --wasmFile=file`);
```
mikedotexe, line 33: @willemneal this PR is awesome!!1 you rock.
Even wrote a test, too. 🤩 

The only thing I think we could improve is this Error. This will give us kind of a nasty stacktrace:

Screen Shot 2020-12-03 at 9 27 28 PM

If we change this to:

```
console.warn(`Multiple Wasm files found:\n\t• ${res.join('\n\t• ')}\n\nPlease pass one with --wasmFile=file`);
process.exit(1);
```

Then we get:

Screen Shot 2020-12-03 at 9 29 50 PM

Whaddayathink?

chadoh: Probably should use console.error instead of console.warn, but otherwise I like your suggestion to stop using 'throw', Mike

utils/find-wasm.js
chadoh: My suggestion is not to change the default behavior, only to remove the explicit default in the code and to document the truer behavior of this improved interface.

utils/find-wasm.js
willemneal: That's weird because I wrote the second find function that catches the error and exits 1.  I am renaming the `findWasmFileSafe` to `findWasmFileUnsafe`, since it is the one that throws errors.  Running locally with the debugger I am realizing that the error jumps above my try catch to the caller of the safe version.

willemneal: I see my issue.  It's async.  So I need to await the result to get the value, otherwise it gets back the promise which then the caller handles.

mikedotexe (approved):

utils/find-wasm.js
mikedotexe: I think Chad makes a good point that we can remove the `default` key and change the help text. I didn't realize this would be backwards-compatible. Good find!

chadoh: Note that we would want to remove the `default` and update the `desc` for both `deploy` and `dev-deploy`. Here's the `desc` I recommend for each:

* `deploy`: File to deploy. Default: "./out/main.wasm" or "**/release/*.wasm"
* `dev-deploy`: File to deploy. Default: "./out/main.wasm" or "**/debug/*.wasm"

vgrichina:
Somewhat related question: how easy it is to add detection whether project is using cargo or asb? For analytics purposes: that's gonna be very useful to determine success rate based on language.

volovyk-s:
@willemneal hey, so we are back to this PR :) Do you think you can solve the conflicts and prepare it for the merge? Because changes look cool.

PR #605, opened 2020-11-17
docs: add units for attached tokens & gas

chadoh:
When using `near call --help`, this will now specify that the attached gas value is specified in gas units while attached deposit is specified in NEAR tokens.


mikedotexe (approved):
great idea

PR #613, opened 2020-11-25
Mixpanel/identity management

icerove:
Cli Identity Management + Consent
Goal: 
Get account ID from users who consent so that we can
Get action + error history for partners / those asked for help --- a more efficient way to help with more context
Understand Cli user journey better for improvements
A unique ID for future full developer journey mapping connecting website, docs, examples, wallets etc.

Implementation Part 1: Consent function

- [x] Add “consent to expose ID to NEAR” in the current AskForConsent Function

Implementation Part 2: Account ID auto storing command
If the user opt in to our tracking, we will only implement tracking code to extract/store their account ID in the commands below: 

- [x] Login <Start + Success>
- [x] NEAR Create Account(wait for discussion to get master or new account)
- [x] Dev Deploy
- [x] Deploy
- [x] Shell Account Status Lookup Started (NEAR State)
- [x] NEAR repl

### DIFF
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -36,7 +36,7 @@ module.exports = {
 };
 
 async function devDeploy(options) {
-    await eventtracking.askForOptInAndAccountID(options);
+    await eventtracking.askForConsentIfNeeded(options);
     const { nodeUrl, helperUrl, masterAccount, wasmFile } = options;
     if (!helperUrl && !masterAccount) {
         throw new Error('Cannot create account as neither helperUrl nor masterAccount is specified in config for current NODE_ENV (see src/config.js)');
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -36,7 +36,7 @@ module.exports = {
 };
 
 async function devDeploy(options) {
-    await eventtracking.askForConsentIfNeeded(options);
+    await eventtracking.askForOptInAndAccountID(options);
     const { nodeUrl, helperUrl, masterAccount, wasmFile } = options;
     if (!helperUrl && !masterAccount) {
         throw new Error('Cannot create account as neither helperUrl nor masterAccount is specified in config for current NODE_ENV (see src/config.js)');
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -1,42 +1,48 @@
 const KeyPair = require('near-api-js').KeyPair;
 const exitOnError = require('../utils/exit-on-error');
 const implicitAccountId = require('../utils/implicit-accountid');
+const connect = require('../utils/connect');
+const eventtracking = require('../utils/eventtracking');
 
 module.exports = {
     command: 'generate-key [account-id]',
     desc: 'generate key or show key from Ledger',
     builder: (yargs) => yargs,
-    handler: exitOnError(async (argv) => {
-        let near = await require('../utils/connect')(argv);
-
-        if (argv.usingLedger) {
-            if (argv.accountId) {
-                console.log('WARN: Account id is provided but ignored in case of using Ledger.');
-            }
-            const publicKey = await argv.signer.getPublicKey();
-            // NOTE: Command above already prints public key.
-            console.log(`Implicit account: ${implicitAccountId(publicKey.toString())}`);
-            // TODO: query all accounts with this public key here.
-            // TODO: check if implicit account exist, and if the key doen't match already.
-            return;
-        }
-
-        const { deps: { keyStore } } = near.config;
-        const existingKey = await keyStore.getKey(argv.networkId, argv.accountId);
-        if (existingKey) {
-            console.log(`Account has existing key pair with ${existingKey.publicKey} public key`);
-            return;
-        }
-
-        // If key doesn't exist, create one and store in the keyStore.
-        // Otherwise, it's expected that both key and accountId are already provided in arguments.
-        if (!argv.publicKey) {
-            const keyPair = KeyPair.fromRandom('ed25519');
-            argv.publicKey = keyPair.publicKey.toString();
-            argv.accountId = argv.accountId || implicitAccountId(argv.publicKey);
-            await keyStore.setKey(argv.networkId, argv.accountId, keyPair);
-        }
-            
-        console.log(`Key pair with ${argv.publicKey} public key for an account "${argv.accountId}"`);
-    })
+    handler: exitOnError(generateKey)
 };
+
+async function generateKey(options) {
+    await eventtracking.askForId(options);
+    const near = await connect(options);
+
+    if (options.usingLedger) {
+        if (options.accountId) {
+            console.log('WARN: Account id is provided but ignored in case of using Ledger.');
+        }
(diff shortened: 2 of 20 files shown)


index.js
```diff
@@ -30,6 +30,7 @@ exports.clean = async function () {
 };
 
 exports.deploy = async function (options) {
+    await eventtracking.askForOptInAndAccountID(options);
```
vgrichina, line 33: this is not necessarily called by user, actually more likely to be used in some deploy script on CI, etc

test/test_account_operations.sh
```diff
@@ -9,7 +9,7 @@ echo Create account
 ../bin/near create-account $testaccount
 
 echo Get account state
-RESULT=$(../bin/near state $testaccount -v | ../node_modules/.bin/strip-ansi)
+RESULT=$(yes | ../bin/near state $testaccount -v | ../node_modules/.bin/strip-ansi)
```
vgrichina, line 12: shouldn't require user interaction, i.e. no `yes |`

test/test_generate_key.sh
```diff
@@ -5,7 +5,7 @@ KEY_FILE=~/.near-credentials/$NODE_ENV/generate-key-test.json
 rm -f "$KEY_FILE"
 echo "Testing generating-key: new key"
 
-RESULT=$(./bin/near generate-key generate-key-test --networkId $NODE_ENV -v)
+RESULT=$(yes |./bin/near generate-key generate-key-test --networkId $NODE_ENV -v)
```
vgrichina, line 8: shouldn't require user interaction, i.e. no `yes |`

utils/eventtracking.js
```diff
@@ -109,7 +122,58 @@ const getEventTrackingConsent = async () => {
     }
 };
 
-const askForConsentIfNeeded = async (options) => {
+const getIdTrackingConsent = async () => {
```
vgrichina, line 125: I think we should only ask people one question, i.e. it's enough to just add text in question above in `getEventTrackingConsent`

utils/eventtracking.js
```diff
@@ -109,7 +122,58 @@ const getEventTrackingConsent = async () => {
...
+    const rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+    });
+    try {
+        for (let attempts = 0; attempts < 10; attempts++) {
+            const answer = await new Promise((resolve) => {
+                rl.question(
+                    chalk`We would like to help with your development journey with NEAR.` +
+                        chalk` We will ask you to share your account Id while using command. ` +
```
vgrichina, line 135: nit: I think we usually spell it as `account ID`, e.g. https://docs.near.org/docs/concepts/account#account-id-rules

vgrichina: I think we also need to add a text to explain to user that this data is already generally on chain. Something like:

`Note that your account ID and all associated on-chain transactions are already being recorded on public blockchain.`

Maybe @kcole16 @corwinharrell can help with text here. Basically we should try to avoid surprises.

utils/eventtracking.js
```diff
@@ -109,7 +122,58 @@ const getEventTrackingConsent = async () => {
...
+
+const askForId = async (options) => {
+    const shellSettings = settings.getShellSettings();
+    if(shouldTrackID(shellSettings)){
+        const id = isGitPod() ? getGitPodUserHash() : shellSettings[TRACKING_SESSION_ID_KEY];
+        await Promise.all([
+            mixpanel.alias(options.accountId, id),
+            mixpanel.people.set(id, {account_id: options.accountId})
+        ]);
+    }else if(shouldNotTrackID(shellSettings)){
```
vgrichina, line 167: nit, please format like in most other places:

```
} else if (shouldNotTrackID(shellSettings)) {
```

utils/eventtracking.js
```diff
@@ -109,7 +122,58 @@ const getEventTrackingConsent = async () => {
...
+        return;
+    }
```
vgrichina, line 169: nit:

```
} else {
```

commands/generate-key.js
```diff
@@ -1,45 +1,50 @@
 const KeyPair = require('near-api-js').KeyPair;
```
vgrichina, line 1: changes in this file seem to be unrelated to the topic of PR, best to revert them

commands/dev-deploy.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
 };
 
 async function devDeploy(options) {
-    await eventtracking.askForConsentIfNeeded(options);
```
vgrichina, line 39: I'd prefer if we keep `askForConsentIfNeeded` name as it is open ended and allows to add more questions if needed. We don't want to rename it to `askForOptInAndAccountIDAndGeoLocation` next time we want to collect some other data.

vgrichina (changes requested):
thanks, moving in right direction and good catch with `near repl`

there are few changes needed however

index.js
```diff
@@ -175,6 +176,7 @@ exports.login = async function (options) {
 };
 
 exports.viewAccount = async function (options) {
+    await eventtracking.askForOptInAndAccountID(options);
```
vgrichina, line 179: this can be used in scripts, not necessarily user-interactive

utils/eventtracking.js
icerove: I do not think so, one question is to generate session id and another one is to track account id which is not the same. @vgrichina

utils/eventtracking.js
icerove: I added

index.js
icerove: @vgrichina what do you mean using in scripts?

vgrichina: I mean that I can write scripts relying on `near-cli` (like e.g. we have scripts in projects created by create-near-app to use `near deploy`, etc).

E.g. I can try to get formatted amount from this call:

```
[I] ➜ near state account.testnet | grep formattedAmount
  formattedAmount: '328.96507017987727742332'
```

this kind of usage breaks if you have a prompt to user. 

We have few commands where user input is expected in any case (like `near login)` – so these can be used to ask user. Commands which don't necessary require user interaction shouldn't be interfered with. Somebody might use them in CI environment where there is no user to prompt. `near deploy` is quite likely to be used in CI for example.

vgrichina: @icerove like e.g. we can see this problem in action on near-cli CI:
https://travis-ci.com/github/near/near-cli/jobs/457657341

it basically hangs on `create-account` call

test/test_account_operations.sh
vgrichina: this is not fixed, why it is resolved?

vgrichina: my bad actually, looks like it is fixed here. But tests don't pass on CI because we still ask for consent in commands which are used in scripts.

index.js
icerove: the function `askForConsentIfNeeded` is asking questions to user instead of let user typing in some account ID or give consent from their side. If I understand correctly, writing scripts is not the case we want for asking process. @vgrichina

PR #614, opened 2020-11-27
Prototype base64-support

vgrichina:


commands/call.js
```diff
@@ -17,8 +17,13 @@ module.exports = {
             type: 'string',
             default: '0'
         })
+        .option('base64',  {
+            desc: 'Treat arguments as base64-encoded BLOB.',
```
chadoh, line 21: I'd rather not add a new argument, if we can avoid it. Why not inspect `args` to figure out what it is?

* if absent, empty string, or starts with `{`, then JSON
* else, try to decode as base64
  * if this causes an error, throw an error explaining the parsing rules

chadoh (changes requested):

commands/call.js
vgrichina: @chadoh I am not sure about this. Maybe we actually want to pass UTF-8 string verbatim when it's not JSON.

chadoh: What do you mean, "pass it verbatim"?

vgrichina: @chadoh I mean that nobody prevents you from implementing your contract in a way where expected arg is just string (vs JSON). Like you pass string `foobar` directly as args instead of passing `{"argumentName":"foobar"}`

chadoh (approved):

djsatok:
This is no longer needed for the bridge, I believe we can close this. @vgrichina

vgrichina:
> This is no longer needed for the bridge, I believe we can close this. @vgrichina

Do you think nobody else would need to pass binary args?

djsatok:
> > This is no longer needed for the bridge, I believe we can close this. @vgrichina
> 
> Do you think nobody else would need to pass binary args?

Potentially need. Do you want to implement anything in addition what is done here? Should we name a PR not a draft?

vgrichina:
we can merge if there are no objections

PR #626, opened 2020-12-11
Add evm-dev-init command to create test accounts

mikedotexe:
Well after looking into proper subcommands in yargs, it ended up not being what's going to work for us. 
Some reading:
https://stackoverflow.com/questions/59894389/how-to-implement-multiple-subcommands-with-yargs
https://laurieontech.com/posts/yargs/
https://github.com/yargs/yargs/blob/master/docs/advanced.md#variadic-positional-arguments

It really doesn't work well with the `--help` flag. I think yargs is essentially not good at having subcommands that take different flags.
I'd like to focus on getting the network config in near-web3-provider, so I'm going to put a fork in this.
Arto, I know you're working on more difficult tasks, don't feel pressure to take your time. This one is super straight forward and I'm happy to merge it after some cursory eyes.

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -246,6 +246,7 @@ yargs // eslint-disable-line
     .command(require('../commands/validators'))
     .command(require('../commands/proposals'))
     .command(require('../commands/evm-call'))
+    .command(require('../commands/evm-dev-init'))
     .command(require('../commands/evm-view'))
     .config(config)
     .alias({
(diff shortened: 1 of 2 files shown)


artob (approved):
LGTM :+1: We can iterate on this later, if need be.

artob:
Just recording here that release [1.5.0](https://github.com/near/near-cli/releases/tag/v1.5.0) contains these goodies.

PR #634, opened 2020-12-16
include keyPath when instantiating NearProvider with evm-call

mikedotexe:
1-liner addressing https://github.com/near/near-web3-provider/issues/68


artob (commented):
LGTM :+1:

PR #684, opened 2021-03-03
View state

vgrichina:


volovyk-s (approved):

commands/view-state.js
```diff
@@ -0,0 +1,45 @@
+const exitOnError = require('../utils/exit-on-error');
+const connect = require('../utils/connect');
+const { formatResponse } = require('../utils/inspect-response');
+
+module.exports = {
+    command: 'view-state <account-id> [prefix]',
```
volovyk-s, line 6: Semantically it's the same as `state`. Should we rename it to `view-contract state`, or simply `contract-state`? @mikedotexe @evgenykuzyakov @chadoh  need your opinion here.

chadoh: Once we have the namespaces proposed in https://github.com/near/NEPs/pull/31, I would advocate for `near contract state`. But I doubt it makes sense to do that for now. 

In the meantime, I don't have a strong preference. `near view-state some-contract.near` seems fine to me.

vgrichina: I'll just roll as is then and makes sense to me to change to namespaces long term.

PR #709, opened 2021-04-08
deprecate default folder

thisisjoshford:
Currently, there are two directories for `testnet` credentials.
  - `default`
  - `testnet`

This can be confusing as `create-near-app` and other example apps store access keys in `testnet` while `near-cli` stores access keys in `default`. Users report running into errors when trying to run `near-cli` commands with credentials in the wrong location. (see linked issue below)

This PR deprecates the `default` directory changing it to `testnet` and prompts users of this change. It also checks the `default` directory for the credentials (if not found in `testnet` ) and asks users if they would like us to copy their keys to the new location. This leaves two copies of an account's access keys to avoid breaking changes until we decide to permanently remove `default`.

Fixes https://github.com/near/near-cli/issues/677

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ exports.clean = async function () {
 };
 
 exports.deploy = async function (options) {
-    await checkCredentials(options.accountId, options.networkId, options.keyStore);
+    if(options.networkId === 'default' | 'testnet') await checkCredentials(options.accountId, options.networkId, options.keyStore);
     console.log(
         `Starting deployment. Account id: ${options.accountId}, node: ${options.nodeUrl}, helper: ${options.helperUrl}, file: ${options.wasmFile}`);
 
@@ -190,7 +190,7 @@ exports.viewAccount = async function (options) {
 };
 
 exports.deleteAccount = async function (options) {
-    await checkCredentials(options.accountId, options.networkId, options.keyStore);
+    if(options.networkId === 'default' | 'testnet') await checkCredentials(options.accountId, options.networkId, options.keyStore);
     console.log(
         `Deleting account. Account id: ${options.accountId}, node: ${options.nodeUrl}, helper: ${options.helperUrl}, beneficiary: ${options.beneficiaryId}`);
     const near = await connect(options);
@@ -209,7 +209,7 @@ exports.keys = async function (options) {
 };
 
 exports.sendMoney = async function (options) {
-    await checkCredentials(options.sender, options.networkId, options.keyStore);
+    if(options.networkId === 'default' | 'testnet') await checkCredentials(options.sender, options.networkId, options.keyStore);
     console.log(`Sending ${options.amount} NEAR to ${options.receiver} from ${options.sender}`);
     const near = await connect(options);
     const account = await near.account(options.sender);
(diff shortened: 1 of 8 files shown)


thisisjoshford:
need to fix CI tests

index.js
```diff
@@ -64,12 +66,15 @@ exports.deploy = async function (options) {
 };
 
 exports.callViewFunction = async function (options) {
+    checkCredentials(options);
```
mikedotexe, line 69: Don't need creds for a view function, eh?

mikedotexe:
Yeah looks like `test_account_creation.sh` behaves differently now and will need to be addressed.

index.js
thisisjoshford: yep! forgot to remove this :)

index.js
```diff
@@ -212,6 +218,7 @@ exports.sendMoney = async function (options) {
 };
 
 exports.stake = async function (options) {
+    if(options.networkId === 'default' | 'testnet') await checkCredentials(options.accountId, options.networkId, options.keyStore);
```
mehtaphysical, line 221: this should be `||`

utils/check-credentials.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = async function checkCredentials(
+    accountId,
+    networkId,
+    keyStore
+) {
+    if (!(await keyStore.getKey(networkId, accountId))) {
+        console.log(
+            `Unable to find [ ${networkId} ] credentials for [ ${accountId} ]`
+        );
+        const defaultKeyPair = await keyStore.getKey('default', accountId);
```
mehtaphysical, line 12: since `networkId` can be either `testnet` or `default` should you check if the key is in the other location? like

```js
const otherNetworkId = networkId === 'default ? 'testnet' : 'default';
const defaultKeyPair = await keyStore.getKey(otherNetworkId, accountId);
```

utils/check-credentials.js
```diff
@@ -0,0 +1,26 @@
+const { askYesNoQuestion } = require('./readline');
+
+module.exports = async function checkCredentials(
+    accountId,
+    networkId,
+    keyStore
+) {
```
mehtaphysical, line 7: this should only run if the networkId is either `testnet` or `default`. maybe add a check:

`if(networkId !== 'testnet' || networkId !== 'default') return;`

That way you break out of these checks early if, say, the networkId is mainnet.

index.js
thisisjoshford: removed this if statement as it is now checking in the function itself

index.js
```diff
@@ -70,6 +72,8 @@ exports.callViewFunction = async function (options) {
     console.log(inspectResponse.formatResponse(await account.viewFunction(options.contractName, options.methodName, JSON.parse(options.args || '{}'))));
 };
 
+
```
mehtaphysical, line 75: Extra whitespace

utils/check-credentials.js
```diff
@@ -0,0 +1,27 @@
...
+            `Unable to find [ ${networkId} ] credentials for [ ${accountId} ]`
+        );
+        const defaultKeyPair = await keyStore.getKey('default', accountId);
+        if (defaultKeyPair) {
+            const answer = await askYesNoQuestion(
+                'Key found in deprecated [ default ] folder. Would you like to copy your key to [ testnet ] folder? (y/n): ',
+                false
+            );
+            if (answer) {
+                console.log('Copying key and retrying transaction...');
```
mehtaphysical, line 20: It’s not really retrying the transaction.

maybe, copying key and attempting transaction?

utils/check-credentials.js
```diff
@@ -0,0 +1,27 @@
...
+                await keyStore.setKey('testnet', accountId, defaultKeyPair);
+            } else {
+                throw 'Please relocate credentials found in [ default ] directory to [ testnet ] directory.';
+            }
+        } else throw 'error';
```
mehtaphysical, line 25: Should have a message to login probably

index.js
mikedotexe: my man. I wasn't gonna say anything 🔍

mikedotexe (approved):
This is looking good, I'll let you and Ryan hash out the details I see that are unresolved. You may fire when ready after consensus is reached between you two.

mehtaphysical (approved):

PR #713, opened 2021-04-15
Check if recipient account exists

ilyar:


vgrichina:
@ilyar what happens without this check, does `sendMoney` fail to throw error about not existing account?

vgrichina:
> @ilyar what happens without this check, does `sendMoney` fail to throw error about not existing account?

nevermind, looks like it creates transaction in that case https://explorer.testnet.near.org/transactions/6zeJEyBCr4hHEVRJuX125p1Yhc7tnD1vs9Jxh7xGSr1s

Issue #666, opened 2021-02-08
near validators <int> doesn't work


ilblackdragon:
**Describe the bug**
Querying validator information for given block doesn't work, as it doesn't parse correctly.

**To Reproduce**
1. `near validators 29238800`
Observe an issue that it can't parse it from JSON RPC

**Expected behavior**
It should provide validator info for given block

PR #715, opened 2021-04-17
epoch converted to number in 'near validators <epoch>'

volovyk-s:
closing #666

### DIFF
--- a/commands/validators.js
+++ b/commands/validators.js
@@ -7,7 +7,7 @@ module.exports = {
     desc: 'lookup validators for given epoch (or current / next)',
     builder: (yargs) => yargs
         .option('epoch', {
-            desc: 'epoch defined by block number or current / next',
+            desc: 'epoch defined by block number, block hash or current / next keyword',
             type: 'string',
             required: true
         }),
--- a/utils/validators-info.js
+++ b/utils/validators-info.js
@@ -2,19 +2,20 @@ const { validators, utils } = require('near-api-js');
 const BN = require('bn.js');
 const AsciiTable = require('ascii-table');
 
-async function validatorsInfo(near, blockNumber) {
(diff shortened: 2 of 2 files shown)


bowenwang1996 (commented):
I am not sure what this is supposed to do. Epoch id is not an integer and cannot be parsed as such

volovyk-s:
@bowenwang1996 looks like the function parameter should be called `blockNumber` (Docs: https://docs.near.org/docs/api/rpc#detailed-validator-status). And it should be a number, not a string. Thanks for pointing this out!
P.S. Check the attached issue for more context.

vgrichina:
> @bowenwang1996 looks like the function parameter should be called `blockNumber` (Docs: https://docs.near.org/docs/api/rpc#detailed-validator-status). And it should be a number, not a string. Thanks for pointing this out!

@volovyk-s looks like there is opportunity to pass block hash as well – we should support either IMO. 

Note that in case you want to have `yargs` always parse some option into number there is such possibility:
https://github.com/yargs/yargs/blob/master/docs/api.md#numberkey

(won't work for the case when you need to support both block hash and number though)

volovyk-s:
> > @bowenwang1996 looks like the function parameter should be called `blockNumber` (Docs: https://docs.near.org/docs/api/rpc#detailed-validator-status). And it should be a number, not a string. Thanks for pointing this out!
> 
> @volovyk-s looks like there is opportunity to pass block hash as well – we should support passing either IMO.
> 
> Note that in case you want to have `yargs` always parse some option into number there is such possibility:
> https://github.com/yargs/yargs/blob/master/docs/api.md#numberkey
> 
> (won't work for the case when you need to support both block hash and number though)

Updated.
To avoid breaking changes I have not added new options (--blockHash and --blockNumber).

PR #717, opened 2021-04-21
Add archival fallback

ilyar:
I think this is not an ideal solution, but I think it is necessary to provide for the possibility of obtaining archived data.

I would be glad to have any suggestions to make a better solution.
```shell
export NEAR_ENV=mainnet
near tx-status wasmgit.near:JBb2DDe3i1CtBwESisLuhxXkWVZpCKYL4J1AdYwAQPsQ
```
Referenses:

https://stackoverflow.com/q/66839103/3123272
https://stackoverflow.com/q/67125552/3123272


volovyk-s:
@ilyar now we are working on a new version of the CLI that will probably have this functionality. But thank you for the contribution!

PR #725, opened 2021-04-28
Ensure NEAR_CLI_OPTIONS includes explicit data for `log-event`

MaximusHaximus:
Re-enables fields that were no longer tracked as of #699

Includes only 'safe' data in `NEAR_CLI_OPTIONS` that is sent to `log-event.js` via environment variable to avoid passing entire yargs object into the child environment

### DIFF
--- a/utils/exit-on-error.js
+++ b/utils/exit-on-error.js
@@ -1,7 +1,11 @@
 const eventtracking = require('./eventtracking');
 const inspectResponse = require('./inspect-response');
 
-function relevantDataFromCmdlineOpts(options) {
+/* `yargs` puts rich objects on the `options` object, which could include sensitive data (like a keystore!).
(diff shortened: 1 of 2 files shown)


volovyk-s:
@MaximusHaximus 
Looks good. Can we rename `relevantDataFromCmdlineOpts` to something like `getNonPrivateDataFromCmdlineOpts`? It would make more sense and prevent the addition of sensible information in the future.

MaximusHaximus:
@volovyk-s :+1: Good idea.  I renamed it as you suggested, and also put a comment there describing how things you might not expect to be there can end up on `options` due to `yargs` behavior :)

volovyk-s (approved):

volovyk-s:
Thanks, @MaximusHaximus !

PR #753, opened 2021-05-12
Generate key for a given HD path improvements

volovyk-s:
Details in: https://github.com/near/near-cli/issues/567

Example:
`serhii@yoga ~/P/N/near-cli (generate-key)> ./bin/near generate-key --useLedgerKey="44'/397'/0'/0'/2'"`
`Make sure to connect your Ledger and open NEAR app`
`Please, confirm on the Ledger receiveng of the public key for HD path 44'/397'/0'/0'/2'`
`Waiting for confirmation on Ledger...`
`Approved`
`Using public key: ed25519:8dxydfuh9NphxaJfavBNPaTJZZkAroRPopbBh7JNzdky`
`Please, confirm that this key is the one that is displayed on the Ledger screen now`
`Waiting for confirmation on Ledger...`
`Approved`
`Using public key: ed25519:8dxydfuh9NphxaJfavBNPaTJZZkAroRPopbBh7JNzdky`
`Implicit account: 717a02a63b986a5ea6fccd8f191ff5087f29809e242e6b3f3fad9f4c03720b5e`

### DIFF
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -15,8 +15,9 @@ module.exports = {
             }
             console.log(`Please, confirm on the Ledger receiveng of the public key for HD path ${argv.useLedgerKey}`);
             const publicKey = await argv.signer.getPublicKey(false);
-            console.log("Please confirm that this key is the one that is displayed on the Ledger screen");
-            await argv.signer.getPublicKey(false, true);
+            if (!publicKey) return;
+            console.log("Please, confirm that this key is the one that is displayed on the Ledger screen now");
+            if (!await argv.signer.getPublicKey(false)) return;
             console.log(`Implicit account: ${implicitAccountId(publicKey.toString())}`);
             // TODO: query all accounts with this public key here.
             // TODO: check if implicit account exist, and if the key doen't match already.
@@ -41,7 +42,7 @@ module.exports = {
             const seededKeyPair = await argv.signer.keyStore.getKey(argv.networkId, argv.accountId);
             await keyStore.setKey(argv.networkId, argv.accountId, seededKeyPair);
         }
-            
+
         console.log(`Key pair with ${argv.publicKey} public key for an account "${argv.accountId}"`);
     })
 };
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -7,7 +7,7 @@ module.exports = {
     desc: 'generate key or show key from Ledger',
     builder: (yargs) => yargs
         .option('yolo', {
-            description: 'Do not ask for extra confitmation when using Ledger',
+            description: 'Do not ask for extra confirmation when using Ledger',
             type: 'boolean',
         }),
     handler: exitOnError(async (argv) => {
@@ -17,7 +17,7 @@ module.exports = {
             if (argv.accountId) {
                 console.log('WARN: Account id is provided but ignored in case of using Ledger.');
             }
-            console.log(`Please, confirm on the Ledger receiveng of the public key for HD path ${argv.useLedgerKey}`);
+            console.log(`Please, confirm on the Ledger receiving the public key for HD path ${argv.useLedgerKey}`);
             const publicKey = await argv.signer.getPublicKey({ enableCaching: false });
             if (!publicKey) {
                 return;
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -25,8 +25,8 @@ module.exports = {
             if (!argv.yolo) {
                 console.log('Please, confirm that this key is the one that is displayed on the Ledger screen now');
                 if (!await argv.signer.getPublicKey({ enableCaching: false })) {
-                    return
-                };
+                    return;
+                }
             }
             console.log(`Implicit account: ${implicitAccountId(publicKey.toString())}`);
             // TODO: query all accounts with this public key here.
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -5,7 +5,11 @@ const implicitAccountId = require('../utils/implicit-accountid');
 module.exports = {
     command: 'generate-key [account-id]',
     desc: 'generate key or show key from Ledger',
-    builder: (yargs) => yargs,
+    builder: (yargs) => yargs
+        .option('yolo', {
+            description: 'Do not ask for extra confitmation when using Ledger',
+            type: 'boolean',
+        }),
     handler: exitOnError(async (argv) => {
         let near = await require('../utils/connect')(argv);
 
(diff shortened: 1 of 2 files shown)


commands/generate-key.js
```diff
@@ -13,8 +13,11 @@ module.exports = {
             if (argv.accountId) {
                 console.log('WARN: Account id is provided but ignored in case of using Ledger.');
             }
-            const publicKey = await argv.signer.getPublicKey();
-            // NOTE: Command above already prints public key.
+            console.log(`Please, confirm on the Ledger receiveng of the public key for HD path ${argv.useLedgerKey}`);
+            const publicKey = await argv.signer.getPublicKey(false);
+            if (!publicKey) return;
```
vgrichina, line 18: nit: best to use `{}` in any case to avoid confusion

middleware/ledger.js
```diff
@@ -16,23 +16,32 @@ module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, n
     const client = await createClient(transport);
 
     let cachedPublicKeys = {};
-    async function getPublicKeyForPath(hdKeyPath) {
+    async function getPublicKeyForPath(hdKeyPath, enableCashing = true) {
```
vgrichina, line 19: `enableCaching`

middleware/ledger.js
```diff
@@ -16,23 +16,32 @@ module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, n
...
+        let rawPublicKey = '';
+        try {
+            rawPublicKey = await client.getPublicKey(false, true);
+            console.log('Approved');
+        } catch (e) {
+            if (e.statusText === 'CONDITIONS_OF_USE_NOT_SATISFIED') {
+                console.log('Rejected from the Ledger ');
+                return;
+            }
+            throw (e);
```
vgrichina, line 34: nit: no `()` needed;

middleware/ledger.js
```diff
@@ -16,23 +16,32 @@ module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, n
...
+    async function getPublicKeyForPath(hdKeyPath, enableCashing = true) {
         // NOTE: Public key is cached to avoid confirming on Ledger multiple times
-        if (cachedPublicKeys[ledgerKeyPath]) {
+        if (cachedPublicKeys[ledgerKeyPath] && enableCashing) {
             return cachedPublicKeys[hdKeyPath];
         }
-
         console.log('Waiting for confirmation on Ledger...');
-        const rawPublicKey = await client.getPublicKey(hdKeyPath);
+        let rawPublicKey = '';
```
vgrichina, line 25: nit: why initialise to `''` if this value is never used?

middleware/ledger.js
```diff
@@ -16,23 +16,32 @@ module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, n
...
+        try {
+            rawPublicKey = await client.getPublicKey(false, true);
```
vgrichina, line 27: nit: very unclear what `false` and `true` use here. Best to pass named variables in such case

vgrichina: https://softwareengineering.stackexchange.com/a/161974

commands/generate-key.js
```diff
@@ -13,8 +13,11 @@ module.exports = {
...
+            console.log('Please, confirm that this key is the one that is displayed on the Ledger screen now');
+            if (!await argv.signer.getPublicKey(false)) return;
```
vgrichina, line 20: not sure I understand why we call `signer.getPublicKey(false)` 2 times in a row

volovyk-s: @vgrichina this is the point of this PR. We want to give users the opportunity to compare what is printed in the terminal and what is shown on the ledger screen. That is why we need to call `getPublicKey` two times. More details in the attached issue.

vgrichina: So user basically would have to approve the key twice on Ledger, correct?

I think it's a bit suboptimal, we probably need to change app itself to work differently. For example on first call ask user to confirm displayed HD path and return public key and on a second call just ask user to confirmed displayed public key.
So basically have 2 diff APIs on Ledger:
1) show HD path, return public key if confirmed
2) display public key for HS path (just for user to confirm it's the same)

Above not blocking this PR however. Maybe makes sense to add `--yolo` option which won't ask for extra confirmation (useful during development for sure and main users are developers here).

middleware/ledger.js
```diff
@@ -16,34 +16,52 @@ module.exports = async function useLedgerSigner({ useLedgerKey: ledgerKeyPath, n
...
+        } catch (e) {
+            if (e.statusText === 'CONDITIONS_OF_USE_NOT_SATISFIED') {
+                console.log('Rejected from the Ledger ');
+                return;
+            }
+            throw e;
+        }
         const publicKey = new PublicKey({ keyType: KeyType.ED25519, data: rawPublicKey });
-        cachedPublicKeys[hdKeyPath] = publicKey;
+        if (enableCaching) { cachedPublicKeys[hdKeyPath] = publicKey; }
```
vgrichina, line 37: nit: sorry for confusion, I meant always use `{}` with `if` like this:

```
if (enableCaching) {
  cachedPublicKeys[hdKeyPath] = publicKey;
}
````

File: commands/generate-key.js (hunk 6 above)
Comment (volovyk-s) [reply commands/generate-key.js]: @vgrichina I have added `--yolo` option and created a separate issue for Ledger here: https://github.com/near/near-ledger-js/issues/42
I would say, that the only way to be sure in you PK and HD path is to show them on the screen at the same time. Two times, since we still need to compare them. In this case, we will need only 1 function that will show the next info: `HD path: path PK: public key` -> [OK]

Comment (vgrichina) [reply commands/generate-key.js]: so there are 2 aspects here:
1) you want to avoid using wrong public key somewhere because near-cli is compromised (this one you already solved)
2) you want to avoid random apps getting public key from Ledger without user consent (that's why you kinda have to ask on Ledger before near-cli displays key)

**Review by vgrichina — APPROVED** [review]

File: commands/generate-key.js [hunk 8]
```diff
@@ -5,16 +5,29 @@ const implicitAccountId = require('../utils/implicit-accountid');
 module.exports = {
     command: 'generate-key [account-id]',
     desc: 'generate key or show key from Ledger',
-    builder: (yargs) => yargs,
+    builder: (yargs) => yargs
+        .option('yolo', {
+            description: 'Do not ask for extra confitmation when using Ledger',
```
mehtaphysical, line 10:
```suggestion
            description: 'Do not ask for extra confirmation when using Ledger',
```

commands/generate-key.js
```diff
@@ -5,16 +5,29 @@ const implicitAccountId = require('../utils/implicit-accountid');
 module.exports = {
     command: 'generate-key [account-id]',
     desc: 'generate key or show key from Ledger',
-    builder: (yargs) => yargs,
+    builder: (yargs) => yargs
+        .option('yolo', {
```
mehtaphysical, line 9: lol :)

commands/generate-key.js
```diff
@@ -5,16 +5,29 @@ const implicitAccountId = require('../utils/implicit-accountid');
...
     handler: exitOnError(async (argv) => {
         let near = await require('../utils/connect')(argv);
 
         if (argv.usingLedger) {
             if (argv.accountId) {
                 console.log('WARN: Account id is provided but ignored in case of using Ledger.');
             }
-            const publicKey = await argv.signer.getPublicKey();
-            // NOTE: Command above already prints public key.
+            console.log(`Please, confirm on the Ledger receiveng of the public key for HD path ${argv.useLedgerKey}`);
```
mehtaphysical, line 20:
```suggestion
            console.log(`Please, confirm on the Ledger receiving the public key for HD path ${argv.useLedgerKey}`);
```

mehtaphysical (commented):
Looks good. Just a few typos

mehtaphysical (approved):

PR #765, opened 2021-05-20
`amount` option renamed to `deposit` (with alias)

volovyk-s:
I think that `deposit` is a better word for a deposit :)

Options list was out of date in README

off top:  probably we should delete dead EVM commands in the next release, so basically we can do that in the next PR

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -14,7 +14,7 @@ module.exports = {
             default: '100000000000000'
         })
         .option('deposit', {
-            desc: 'Number of tokens to attach (in NEAR) in a function call',
+            desc: 'Number of tokens to attach (in NEAR) to a function call',
             type: 'string',
             default: '0',
             alias: 'amount'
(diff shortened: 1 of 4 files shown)


commands/call.js
```diff
@@ -13,10 +13,11 @@ module.exports = {
             type: 'string',
             default: '100000000000000'
         })
-        .option('amount', {
-            desc: 'Number of tokens to attach (in NEAR)',
+        .option('deposit', {
+            desc: 'Number of tokens to attach (in NEAR) in a function call',
```
mehtaphysical, line 17:
```suggestion
            desc: 'Number of tokens to attach (in NEAR) to a function call',
```

mehtaphysical (approved):
Looks good. Just one suggestion

commands/call.js
volovyk-s: Thanks!

thisisjoshford (approved):
LGTM. Let me know when you merge and I'll update docs :)

PR #782, opened 2021-06-01
update near call to use object args

### LINKED ISSUE #781: `near call` is using deprecated `functionCall` method
`near call` is currently using deprecated version of `account.functionCall` method that uses positional arguments. We should update this in `near-cli` that passes an object for args.

thisisjoshford:
In `near-api-js`, calling `account.functionCall` using positional args is being deprecated. This PR updates this command to use the newer function call method of passing arguments as an object.

Fixes #781

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -11,28 +11,28 @@ module.exports = {
         .option('gas', {
             desc: 'Max amount of gas this call can use (in gas units)',
             type: 'string',
-            default: '100000000000000',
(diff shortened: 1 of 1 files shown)


thisisjoshford:
> looks good. There are a bunch of whitespace changes though that aren't necessary.

It's what Prettier auto performed on save. Shall I revert the changes @mehtaphysical ?

volovyk-s (approved):

Issue #787, opened 2021-06-05
view-state fails without --finality flag


volovyk-s:
**Describe the bug**
view-state fails without --finality flag

**To Reproduce**
`near view-state <accId>`

**Expected behavior**
Should clarify that finality is required

chadoh:
I'd prefer we default to `--finality final`, rather than always require it.

Are there other commands that currently require this flag?

volovyk-s:
@chadoh no, this is the only one
It can not be a default parameter because it is in conflict with `blockId`.

PR #799, opened 2021-06-16
fix view-state finality flag

ailisp:
I ran into #787 too. finality cannot simply be add as required param, because user need to specify one of blockid and finality and only one. Otherwise near-api-js fails. Also fixed two bugs:
1) passing blockid to near-api-js, which expects block_id and blockId cause crash; 
2) block_id could not take number, because command line arg was also considered as string and near-api-js tried to interpret it as block hash, fix by convert number like string to javascript string.
Result:
```
near-cli (fix-787) bin/near view-state simple-state.abcdef.testnet --blockId 51762759
[
  {
    key: <Buffer 53 54 41 54 45>,
    value: <Buffer 01 00 00 00 0e 00 00 00 61 62 63 64 65 66 2e 74 65 73 74 6e 65 74 05 00 00 00 68 65 6c 6c 6f>
  }
]
near-cli (fix-787) bin/near view-state simple-state.abcdef.testnet --blockId Fg4CVqhu1c1DdC5WegicGsE2NJYUB1K1wrwvvPPF1e4d
[
  {
    key: <Buffer 53 54 41 54 45>,
    value: <Buffer 01 00 00 00 0e 00 00 00 61 62 63 64 65 66 2e 74 65 73 74 6e 65 74 05 00 00 00 68 65 6c 6c 6f>
  }
]
near-cli (fix-787) bin/near view-state simple-state.abcdef.testnet --block-id 51762759 --finality final
Only one of --finality and --blockId can be provided
1|near-cli (fix-787) bin/near view-state simple-state.abcdef.testnet
Must provide either --finality or --blockId
1|near-cli (fix-787) bin/near view-state simple-state.abcdef.testnet --finality final
[
  {
    key: <Buffer 53 54 41 54 45>,
    value: <Buffer 01 00 00 00 0e 00 00 00 61 62 63 64 65 66 2e 74 65 73 74 6e 65 74 05 00 00 00 68 65 6c 6c 6f>
  }
]
```

### DIFF
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -22,7 +22,7 @@ module.exports = {
                 '`final` is for a block that has been validated on at least 66% of the nodes in the network',
             type: 'string',
             choices: ['optimistic', 'final'],
-            
+            default: 'final',
         })
         .option('utf8', {
             desc: 'Decode keys and values as UTF-8 strings',
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -14,15 +14,13 @@ module.exports = {
         .option('block-id', {
             desc: 'The block number OR the block hash (base58-encoded).',
             type: 'string',
-            
-        })
-        .coerce({
-            blockId: (blockId) => {
+            coerce: (blockId) => {
                 if (blockId && !isNaN(blockId)) {
                     return Number(blockId);
                 }
                 return blockId;
             }
+            
         })
         .option('finality', {
             desc: '`optimistic` uses the latest block recorded on the node that responded to your query,\n' + 
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -36,10 +36,10 @@ async function viewState(options) {
     const near = await connect(options);
     const account = await near.account(accountId);
     if (finality && blockId) {
-        console.error("Only one of --finality and --blockId can be provided");
+        console.error('Only one of --finality and --blockId can be provided');
         process.exit(1);
     } else if (!finality && !blockId) {
-        console.error("Must provide either --finality or --blockId");
+        console.error('Must provide either --finality or --blockId');
         process.exit(1);
     }
     // near-api-js takes block_id instead of blockId
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -1,7 +1,6 @@
 const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const { formatResponse } = require('../utils/inspect-response');
-const yargs = require('yargs');
 
 module.exports = {
     command: 'view-state <account-id> [prefix]',
@@ -29,7 +28,7 @@ module.exports = {
             default: false
         })
         .conflicts('block-id', 'finality')
-        .check((argv, _) => {
+        .check((argv) => {
             if (!argv.finality && !argv.blockId) {
                 throw new Error('Must provide either --finality or --blockId');
             }
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -22,7 +22,6 @@ module.exports = {
                 '`final` is for a block that has been validated on at least 66% of the nodes in the network',
             type: 'string',
             choices: ['optimistic', 'final'],
-            default: 'final',
         })
         .option('utf8', {
             desc: 'Decode keys and values as UTF-8 strings',
@@ -36,8 +35,21 @@ async function viewState(options) {
     const { accountId, prefix, finality, blockId, utf8 } = options;
     const near = await connect(options);
     const account = await near.account(accountId);
-
-    let state = await account.viewState(prefix, { blockId, finality });
+    if (finality && blockId) {
+        console.error("Only one of --finality and --blockId can be provided");
+        process.exit(1);
+    } else if (!finality && !blockId) {
+        console.error("Must provide either --finality or --blockId");
+        process.exit(1);
+    }
+    // near-api-js takes block_id instead of blockId
+    let block_id = blockId;
+    if (blockId && !isNaN(Number(blockId))) {
+        // If block id is a number (still string as command line args), it must be convert to JavaScript Number
+        // for near-api-js.
+        block_id = Number(blockId);
+    }
+    let state = await account.viewState(prefix, { block_id, finality });
     if (utf8) {
         state = state.map(({ key, value}) => ({ key: key.toString('utf-8'), value: value.toString('utf-8') }));
     }
--- a/commands/view-state.js
+++ b/commands/view-state.js
@@ -16,6 +16,14 @@ module.exports = {
             type: 'string',
             
         })
+        .coerce({
+            blockId: (blockId) => {
+                if (blockId && !isNaN(blockId)) {
+                    return Number(blockId);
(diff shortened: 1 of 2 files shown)


ailisp:
We probably should add more ci test, near view-state with given block id was not working and not caught. When adding new command it should come with tests. For exisiting command, I suggest for every crashed case add a test to avoid regression.

volovyk-s:
> We probably should add more ci test, near view-state with given block id was not working and not caught. When adding new command it should come with tests. For exisiting command, I suggest for every crashed case add a test to avoid regression.

I have added two tests for `--finality final` and `--finality optimistic`.
I'm wondering if we can get storage by block or hash, because it becomes inaccessible with time (tell me if I'm wrong).

volovyk-s (approved):

ailisp:
Nice! just studied your test, I think I learned how to do it in future PRs. As for block hash and ID, i agree, old blocks would be GCed and disappeared. Maybe I can give a try of a test that: first fetch latest block height and hash, then run near-view state with this block

ailisp:
@volovyk-s I added some more tests follow your format, by get current block/hash via near state account and near view-state with that block height/hash

volovyk-s:
> @volovyk-s I added some more tests follow your format, by get current block/hash via near state account and near view-state with that block height/hash

Awesome! @thisisjoshford , @mehtaphysical we still need one review ;)

mikedotexe:
> > @volovyk-s I added some more tests follow your format, by get current block/hash via near state account and near view-state with that block height/hash
> 
> Awesome! @thisisjoshford , @mehtaphysical we still need one review ;)

feel free to merge it if it's reviewed and approved, thanks

commands/view-state.js
```diff
@@ -36,8 +35,21 @@ async function viewState(options) {
     const { accountId, prefix, finality, blockId, utf8 } = options;
     const near = await connect(options);
     const account = await near.account(accountId);
-
-    let state = await account.viewState(prefix, { blockId, finality });
+    if (finality && blockId) {
```
MaximusHaximus, line 38: For mutually exclusive options (only one of either finality vs. blockId), we should try using .conflicts() (ref: https://yargs.js.org/docs/#api-reference-conflictsx-y)

commands/view-state.js
```diff
@@ -36,8 +35,21 @@ async function viewState(options) {
...
+        console.error('Only one of --finality and --blockId can be provided');
+        process.exit(1);
+    } else if (!finality && !blockId) {
+        console.error('Must provide either --finality or --blockId');
+        process.exit(1);
+    }
+    // near-api-js takes block_id instead of blockId
+    let block_id = blockId;
+    if (blockId && !isNaN(Number(blockId))) {
```
MaximusHaximus, line 47: For parsing of inputs data, we should use .coerce() (ref: https://yargs.js.org/docs/#coerce) -- failures to coerce arguments into our desired datatype should throw an error rather than exiting the process, so Yargs can generate user-friendly, consistent output.

commands/view-state.js
```diff
@@ -36,8 +35,21 @@ async function viewState(options) {
...
-    let state = await account.viewState(prefix, { blockId, finality });
+    if (finality && blockId) {
+        console.error('Only one of --finality and --blockId can be provided');
+        process.exit(1);
+    } else if (!finality && !blockId) {
+        console.error('Must provide either --finality or --blockId');
+        process.exit(1);
+    }
+    // near-api-js takes block_id instead of blockId
+    let block_id = blockId;
```
MaximusHaximus, line 46: YArgs automatically creates properties on options for both the hypenated and non-hypenated strings for each option -- there's no need to have a local block_id variable here at all, you can just change L35 to reference options.block-id :)

commands/view-state.js
```diff
@@ -36,8 +35,21 @@ async function viewState(options) {
     const { accountId, prefix, finality, blockId, utf8 } = options;
     const near = await connect(options);
     const account = await near.account(accountId);
-
-    let state = await account.viewState(prefix, { blockId, finality });
+    if (finality && blockId) {
+        console.error('Only one of --finality and --blockId can be provided');
+        process.exit(1);
+    } else if (!finality && !blockId) {
+        console.error('Must provide either --finality or --blockId');
```
MaximusHaximus, line 42: For 'we require either one of two options to be includes', use .check() (ref: https://yargs.js.org/docs/#api-reference-checkfn-globaltrue) and throw from the function if the validation fails (instead of process.exit())

MaximusHaximus (changes requested):
Thanks so much for digging in and tackling this problem, @ailisp!

Rather than adding custom-rolled validation in the actual command handlers, we should be validating things _inside yargs_, so that our command handlers are lean and assume that they have been given valid arguments, and we are leveraging the benefits that `yargs` provides us around user feedback and error handling.  I've added some comments in-line, as we're re-implementing some of the core functionality of yargs here with our own code; we should be defining these cases in our builder and option objects so that the output provided to the user is consistent with other validation failures and we're not tying validation logic to our command execution logic

Also, as a general note using process.exit(1) is an antipattern in almost every case, and is usually a code smell that something is not properly encapsulated -- it makes unit testing the code impossible and forces us to make a full external integration test because the entire test framework process would exit when these cases are encountered, making it impossible for us to assert in native JS that it is working how we expect. By using YArgs for these types of things, we let YArgs exit the process if it's the right thing to do, but don't make any assumptions ourselves that when these few lines of code are run that we should exit the process.

commands/view-state.js
MaximusHaximus: Ah, my apologies -- I just realised that this was a snake_case conversion, not a dashed-property :+1: nevermind this one :)

commands/view-state.js
ailisp: Nice! I didn't know this

ailisp:
@MaximusHaximus Ptal, thanks!

commands/view-state.js
```diff
@@ -22,12 +21,26 @@ module.exports = {
...
+        })
+        .conflicts('block-id', 'finality')
+        .check((argv) => {
+            if (!argv.finality && !argv.blockId) {
+                throw new Error('Must provide either --finality or --blockId');
+            }
+            return true;
+        })
+        .coerce({
+            blockId: (blockId) => {
```
MaximusHaximus, line 38: :nit: Not a blocker, but this function can be co-located with the `block-id` option definition (unlike `conflicts` which must be globally defined) -- `coerce` is a valid property that you can provide to the `.option()` call along with stuff like `desc`, `type`, `default`, `required`, etc. :)

MaximusHaximus (approved):
Just one :nit: that isn't a blocker -- but feel free to change it, as it would be a bit nicer if the coercion for `block-id` was defined with the rest of its properties :)

MaximusHaximus:
@allisp Just a little different than that :D I went ahead and added a commit so you can see what I mean rather than have you spend more time messing with it :)

MaximusHaximus:
@ailisp Just a little different than that :D I went ahead and added a commit so you can see what I mean rather than have you spend more time messing with it :)

PR #822, opened 2021-07-20
feat: Add deposit yocto

ChaoticTempest:
Closes https://github.com/near/near-cli/issues/751

This only adds the `depositYocto` option to the call command. 

I didn't add it to other commands since they were still using `amount` as an option. Should we go ahead and change those to `deposit`/`depositYocto` as well?

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -19,6 +19,11 @@ module.exports = {
             default: '0',
             alias: 'amount'
         })
+        .option('depositYocto', {
+            desc: 'Number of tokens to attach (in yocto NEAR) to a function call',
+            type: 'string',
+            default: null,
+        })
         .option('base64', {
             desc: 'Treat arguments as base64-encoded BLOB.',
             type: 'boolean',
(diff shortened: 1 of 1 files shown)


ChaoticTempest:
Also, Travis CI is failing on my usage of null coalescing. What JS version are we using?

commands/call.js
```diff
@@ -19,6 +19,11 @@ module.exports = {
             default: '0',
             alias: 'amount'
         })
+        .option('depositYocto', {
```
volovyk-s, line 22: I don't think that we should add a separate option here.
And we definitely should not use it as a flag.
@mikedotexe I do not understand the motivation in your comment here: https://github.com/near/near-cli/issues/751#issuecomment-882743407

I would say, that we need to design it the next way:
`near call ... ... --deposit 000000001` it would be in N, for backward compatibility + deprecation warning.
`near call ... ... deposit 1N`
`near call ... ... deposit 1yN` for deposit in yoctoNEAR

If we know for sure that `deposit` is always in yoctoNEAR, then let's do it the next way:
`near call ... ... --depositYocto 1` + add deprecation warning for `--deposit`
This one will be less flexible.

+, if we will have `--deposit` and `--depositYocto` at the same time we should not allow to specify them both (there is a conflicting functionality in `yargs`).

@ChaoticTempest please, provide an example of this option usage in the description.

volovyk-s:
> Also, Travis CI is failing on my usage of null coalescing. What JS version are we using?

You should be able to check it somewhere in the CI setup. Do you have an access?

volovyk-s:
> I didn't add it to other commands since they were still using `amount` as an option. Should we go ahead and change those to `deposit`/`depositYocto` as well?

I believe that `amount` options do not always mean `deposit`. But if you thing that it should be renamed somewhere - that would be cool :+1:  :)

ChaoticTempest:
> You should be able to check it somewhere in the CI setup. Do you have an access?

Oh I see now, it's using node js 12

> I believe that `amount` options do not always mean `deposit`. But if you thing that it should be renamed somewhere - that would be cool 👍 :)

Ahh, I must've been mistaken then when I looked at it. No need to change it to `deposit` then

commands/call.js
ChaoticTempest: In the PR description or somewhere else?

volovyk-s: Yes, let's put it there :)

mikedotexe (approved):

PR #830, opened 2021-08-02
Dependabot config

volovyk-s:
`monthly` — runs once each month. This is on the first day of the month.
source: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval

`open-pull-requests-limit` increased to 20 since we can have more than 10 updates in 1 month, and 2 months delay is too much

`dependabot` config is not super-flexible, the only other option to make it less frequent is to whitelist the libraries that we want to update (not the best solution on my opinion).


.github/dependabot.yml
```diff
@@ -3,8 +3,8 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: daily
-  open-pull-requests-limit: 10
+    interval: monthly
+  open-pull-requests-limit: 20
```
chadoh, line 7: If it's only happening monthly, is it better to remove the limit altogether?

chadoh (changes requested):
I like monthly for external packages, but I wonder if it makes sense to update near-cli with updated NEAR packages more quickly, somehow. But maybe we do that with a process outside of Dependabot. What are your thoughts?

Also, see my comment below. Feel free to push back on my points; I only want to at least discuss them prior to merge, but will defer to you if you think the current approach is best.

.github/dependabot.yml
volovyk-s: @chadoh quote from docs:
`open-pull-requests-limit`
`By default, Dependabot opens a maximum of five pull requests for version updates. Once there are five open pull requests, new requests are blocked until you merge or close some of the open requests, after which new pull requests can be opened on subsequent updates. Use open-pull-requests-limit to change this limit. This also provides a simple way to temporarily disable version updates for a package manager.`
`This option has no impact on security updates, which have a separate, internal limit of ten open pull requests.`

I think that it should remain, we can have more than 5 updates in 1 month.

volovyk-s: @chadoh this approach is not documented, but I hope that it will work.
Now we have 2 jobs, 1 will check for updates monthly and another one will do that daily, but it will ignore all dependencies except our won libraries.

chadoh (approved):
Makes sense to me! Let's try it.

mikedotexe (approved):

volovyk-s:
Sorry @chadoh and @mikedotexe , dependabot gives us this error:
`The property '#/updates/1' is a duplicate. Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'`
So it's not possible to have multiple update schedules in one repo.

There is a year-old issue: https://github.com/dependabot/dependabot-core/issues/2390

I will revert it to the previous state, take a look.

MaximusHaximus:
It's critical to understand that `dependabot` is the _same_ bot that creates PRs to bump package versions  _due to security CVE's on the existing package we use_. This means that having a large amount of PRs for trivial, often completely unnecessary (for us) and zero-impact (for us) minor releases for our dependent packages will create a bunch of noise that we need to filter through to identify which PRs were created by dependabot due to verified security issues in our other dependent packages, (you have to open the PR and look at it to tell, and can only see if it is a security fix PR if you have access to `dependabot alerts` in that repository.)

I'd like to step back a little from the configuration changes for dependabot and talk a little about dependency management and the benefits and drawbacks we get from dependabot creating updates regularly for us.

First, let's consider the risk of supply chain attacks by way of dependencies:
A. Security-wise, we really _do **not** want to be on the latest, just released NPM package minor release version_ because this puts us in the early-adopter cohort for that newly published code, which is most likely to be involved in a supply chain NPM hack. This is because, when/if someone compromises an NPM package and they want to catch the most consumers with their new access, they will _almost always_ publish a tiny minor point release, because many projects only auto-update that level due to the possibility of breaking changes in non minor releases that could break their existing codebase's usage patterns for that package.  Also, it makes it less likely that people will notice the hack because point releases can sometimes happen very regularly for some packages, and even the people who own/maintain that package (in the case of open source projects with many contributors) might not remember exactly which point release they are on.
B. Security-wise, bumping package versions is _dangerous_. It should be done _carefully_ and _deliberately_.
- We should be absolutely sure that the newly published package is legitimate; e.g. is there a corresponding commit to the Github repo that is the source for the NPM package?
- Is the code in the NPM package the same as the release in the Github repo, or did someone publish an NPM package that contains code that does not match the Github release?
- We must be confident that the  changeset we're merging on that dependency it is exactly as expected and doesn't contain malicious code or security vulnerabilities that could be exploited.
- This type of review is somewhat less important based on how widely used the package in question is, and how long the release has been 'in the wild', because it gives time for the package consumer community to identify unwanted behaviour or supply chain attacks, but it is _**absolutely essential**_ if we intend to use the standard dependabot strategy of always bumping to the latest minor release for a package, because the code we are upgrading to may have only been installed by a tiny subset of consumers when we upgrade to it.

Second, let's consider what dependency management means for us in particular at the codebase and organization level.
Given (A) above, keeping our dependencies 100% up to date is _extremely expensive_ effort-wise from the engineering/developer resource perspective.
- It requires someone who is _highly proficient_ both with Javascript as a language and with the NPM/github ecosystem to be constantly on top of automatically generated PRs to avoid missing critical security PRs that we should merge ASAP vs. point releases for stuff that isn't even relevant for our use cases.
- It requires that whoever is responsible for bumping these dependencies spends time _regularly_ reviewing the code that will change when the proposed PRs are merged before doing so.
- If most of the PRs created _will not be relevant to us_, (and most of them won't be, in my experience) then we will be taking on a huge amount of developer time reviewing dependencies and updating them for no direct benefit.
- If a point release doesn't touch a feature we rely on, or doesn't add functionality that we find useful, then upgrading to it actually risks us adding a defect to our existing codebase that is part of that changeset, even though merging that changeset doesn't even provide us any benefits other than being able to say 'we are at the latest version of X' -- which isn't always a benefit (see above re: security).
- Given that any change that we merge could include a bug that could impact us.  Do we have extensive integration tests with our dependencies? How will we know if bumping to the latest version of xyz package has caused a bug in our usage of that dependency?  Are we going to have to manually test these changesets? And again, did we gain anything from merging that code update, or did we add a defect by merging code that actually didn't provide any benefits to us from a functional, reliability or security perspective?

Due to the above considerations, I recently switched both `near-wallet` and `near-contract-helper` repositories to _not use dependabot for automatic package dependency PRs _, and _only_ for _security related PRs_.

IMO _if_ we want to use `dependabot` to manage automatic PRs for package version updates, it makes a lot of sense to limit those PRs to only **major** version bumps, which massively limits the chance we get caught in a supply chain compromise, and reduces the amount of effort we spend on dependency management that doesn't have any direct impact on our usage of those dependencies.  Also, it ensures that we don't end up falling behind on breaking changes for packages that we depend on, which could result in a situation where a security patch might _require_ that we upgrade to a new major version, but we can't update to it until we've refactored our code to resolve breaking changes in the major version API interface.  Lastly, it makes sure that we are upgrading to the new major releases while good documentation is still easily accessible for the prior release; many packages don't do a good job of keeping migration instructions around for e.g. 3 major releases ago and it can be tricky to upgrade a package by 2-3 major releases in one go because most instructions won't tell you how to go from v3 to v5...but instead you end up doing all the work to convert from v3 to v4, then from v4 to v5 (for example).

In the Wallet's case, we will do an audit of our dependencies every 6 months or so and create explicit tickets to track updating our dependencies, _when we need them_, which is to say _when upgrading to the new version provides benefits for our uses of those packages_, but any dependabot PRs for security issues will be immediately investigated and merged ASAP if they are relevant/exploitable in our code. Just wanted to put this out there for consideration :)

mikedotexe (approved):

mikedotexe:
> Due to the above considerations, I recently switched both `near-wallet` and `near-contract-helper` repositories to _not use dependabot for automatic package dependency PRs_, and _only_ for _security related PRs_.

How did you do this? I've been clicking around like crazy and can't see. I had tried this before a couple week ago and similarly couldn't find it. 😕

MaximusHaximus:
So, in theory all you need to do is to set `open-pull-requests-limit` to `0` in the dependabot.yml config, and the official documentation indicates that dependabot will still create security PRs with its own _internal limit_ of `10`.

However, I found multiple reports from folks that it doesn't actually work and that the limit also stopped security PRs from being created.  

So somewhere I found an issue in the dependabot repository where folks described that if you set a specific `target-branch` for dependabot.yml and set its configured `open-pull-requests-limit` value to `0`, it would disable the dependency PRs but leave security PR behaviours intact.

So the solution I implemented was to create a `dump` branch and modify `dependabot`'s `npm` package configuration to point at that branch, but set it's value for `open-pull-requests-limit` to 0.  The 'default dependabot config' (which you can't view anywhere that I could find, and that _always_ exists) still applies for the security PRs in this case.

It's _possible_ that you could just set `open-pull-requests-limit` to 0 in the main dependabot.yml config for `npm`, and it might still create security PRs... I did not test this, because I found some people reporting that it didn't work unless they used this workaround... so I went straight to the workaround.

https://github.com/near/near-wallet/blob/master/.github/dependabot.yml is the config we use
And here's a security PR opened this week showing it does open security PRs (https://github.com/near/near-wallet/pull/2030)

volovyk-s:
@MaximusHaximus @mikedotexe I like this strategy, and I agree that usually there are no benefits from upgrading to the last minor version of the package. The downside in this approach is that we need to keep `dump` branch up to date. I hope that they will fix it soon.
@MaximusHaximus Have you subscribed to that issue? Can you send a link? We will need to change it to `master` once it is fixed.

volovyk-s:
@mikedotexe we will need to create a calendar event for dependency updates. Especially if we will use this strategy in other projects.

MaximusHaximus:
> @MaximusHaximus @mikedotexe I like this strategy, and I agree that usually there are no benefits from upgrading to the last minor version of the package. The downside in this approach is that we need to keep `dump` branch up to date. I hope that they will fix it soon.
> @MaximusHaximus Have you subscribed to that issue? Can you send a link? We will need to change it to `master` once it is fixed.

Actually, it doesn't matter what is in the dump branch, because dependabot won't create any PRs against it; the only thing that's important is that the `dump` branch _exists_, I think.  The only reason why we set an explicit `targetBranch` is so that the configuration for version bump packages doesn't apply to the main branch in the repo.  Essentially we've created a brand new configuration for version bump PRs that only ever creates pulls against `dump` branch, but then we set that configuration to never create any PRs by setting it to allow 0 PRs.  The default config 'shadows' this config and applies to security updates, which always apply to the main branch of the repo.  It's a weird thing, but it means we don't need to worry about maintaining `dump` branch :)

PR #836, opened 2021-09-08
support localnet

H34D:
The `local` network has been renamed to `localnet` in `nearup` so it hink it should be reflected here as well

### DIFF
--- a/config.js
+++ b/config.js
@@ -45,8 +45,9 @@ function getConfig(env) {
             helperAccount: 'guildnet',
         };
     case 'local':
+    case 'localnet':
         return {
-            networkId: 'local',
+            networkId: 'localnet',
             nodeUrl: 'http://localhost:3030',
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -45,8 +45,9 @@ function getConfig(env) {
             helperAccount: 'guildnet',
         };
     case 'local':
+    case 'localnet':
         return {
-            networkId: 'local',
+            networkId: 'localnet',
```
volovyk-s, line 50: This will be a breaking change, I would rather avoid it. Do you have a strong motivation to merge it? Can you describe a use case?

H34D: I tested this extensively and it indeed caused issues in other components. I rolled back this change.

volovyk-s: Thanks!

volovyk-s (approved):

PR #862, opened 2021-10-24
Restrict to delete last FullAccess key

TarasenkoFirelabs:
Check does it really last FullAccess key
If user want to delete any other key, will allow this action


volovyk-s:
@TarasenkoFirelabs let's add everything we need in https://github.com/near/near-cli/pull/864

PR #863, opened 2021-10-24
Add ability to generate secp256k1 keys

TarasenkoFirelabs:
solved issue https://github.com/near/near-cli/issues/855

Note that we already have this in the Rust SDK:
https://github.com/near/near-sdk-rs/blob/2e979934cca7c87a5ca5d7dcb62e4b3cbe809afc/near-sdk/src/types/public_key.rs#L10

and

https://github.com/near/near-sdk-rs/blob/1a706b2eb0dd3e96e2ba4f6e74b0ef972c4bbbdf/sys/src/lib.rs#L38

and

https://github.com/near/nearcore/blob/a20c12ec02d7f22550139da38838efd1fba75ede/runtime/near-vm-logic/src/logic.rs#L976

### DIFF
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -10,6 +10,10 @@ module.exports = {
         .option('yolo', {
             description: 'Do not ask for extra confirmation when using Ledger',
             type: 'boolean',
+        })
+        .option('curveType', {
+            description: 'Use curve type ed25519 or secp256k1 for key pair generate',
+            type: 'string',
         }),
     handler: exitOnError(async (argv) => {
         let near = await require('../utils/connect')(argv);


mikedotexe (changes requested):
Similar to how we have the optional `yolo` argument, let's please add the `curveType` so that yargs knows about it.
(Like here: https://github.com/near/near-cli/blob/4cf5e360dc9a87fe5d89bfb56e0ec342c98f2cd3/commands/generate-key.js#L9-L12)

commands/generate-key.js
```diff
@@ -1,3 +1,4 @@
+const keyCurveTypes = require('near-api-js').keyCurveTypes;
```
volovyk-s, line 1: @TarasenkoFirelabs this line gives me undefined.

TarasenkoFirelabs: I am still working onto this task in `near-api-js` repo

volovyk-s: Got it, then it's better to keep this one in a draft. Ping me when it will be ready for review, thanks!

PR #864, opened 2021-10-25
Restrict to delete last full access key promted

TarasenkoFirelabs:
https://github.com/near/near-cli/issues/834

add prompt Y\n if try to delete last full access key

### DIFF
--- a/commands/delete-key.js
+++ b/commands/delete-key.js
@@ -37,8 +37,8 @@ async function deleteAccessKey(options) {
             const result = await account.deleteKey(options.accessKey);
             inspectResponse.prettyPrintResponse(result, options);
         } else {
-            console.log('Deleting key canceled.');
             rl.close();
+            console.log('Deleting key canceled.');
         }
     }
 }
--- a/commands/delete-key.js
+++ b/commands/delete-key.js
@@ -33,13 +33,12 @@ async function deleteAccessKey(options) {
         const question = (str) => new Promise(resolve => rl.question(str, resolve));
         const answer = await question('WARN: you want to remove the last full access key and forgot access to this account [Y/n]? ');
 
-        if(['N', 'n', 'no', 'No', 'NO'].includes(answer)) {
+        if (['YES', 'Yes', 'yes', 'Y', 'y'].includes(answer)) {
+            const result = await account.deleteKey(options.accessKey);
+            inspectResponse.prettyPrintResponse(result, options);
+        } else {
             console.log('Deleting key canceled.');
             rl.close();
-            process.exit(1);
         }
     }
-
-    const result = await account.deleteKey(options.accessKey);
-    inspectResponse.prettyPrintResponse(result, options);
 }
--- a/commands/delete-key.js
+++ b/commands/delete-key.js
@@ -2,7 +2,8 @@ const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
 const inspectResponse = require('../utils/inspect-response');
 const checkCredentials = require('../utils/check-credentials');
-const readline = require('readline');
+const chalk = require('chalk');
+const { askYesNoQuestion } = require('./readline');
 
 module.exports = {
     command: 'delete-key <account-id> <access-key>',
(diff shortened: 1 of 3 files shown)


commands/delete-key.js
```diff
@@ -20,6 +21,25 @@ async function deleteAccessKey(options) {
...
+    if (fullAccessKeys.length === 1 && fullAccessKeys[0].public_key.includes(options.accessKey)){
+        const rl = readline.createInterface({
+            input: process.stdin,
+            output: process.stdout
+        });
+
+        const question = (str) => new Promise(resolve => rl.question(str, resolve));
+        const answer = await question('WARN: you want to remove the last full access key and forgot access to this account [Y/n]? ');
+
+        if(['N', 'n', 'no', 'No', 'NO'].includes(answer)) {
```
volovyk-s, line 36: Thank you @TarasenkoFirelabs! This contribution is super important.
Since we want to ensure that the user will not delete his last key by accident, let's check for `YES`/`Yes`/`Y`/`y` and delete key `ONLY` if these answers are provided.
Also, with this approach we will not use `precess.exit(1)` which is not a good practice.

TarasenkoFirelabs: solved

commands/delete-key.js
```diff
@@ -20,6 +21,24 @@ async function deleteAccessKey(options) {
     console.log(`Deleting key = ${options.accessKey} on ${options.accountId}.`);
     const near = await connect(options);
     const account = await near.account(options.accountId);
-    const result = await account.deleteKey(options.accessKey);
-    inspectResponse.prettyPrintResponse(result, options);
+    let accessKeys = await account.getAccessKeys();
+    let fullAccessKeys = accessKeys.filter(accessKey => accessKey.access_key.permission === 'FullAccess');
```
volovyk-s, line 25: Looks like we will not delete the key if this condition is not met.
It's a critical piece of code. Let's work on it together. Here is a thought about warnings that we should have here:
https://github.com/near/near-cli/issues/834#issuecomment-968970448

commands/delete-key.js
```diff
@@ -20,6 +21,24 @@ async function deleteAccessKey(options) {
...
+        const answer = await question('WARN: you want to remove the last full access key and forgot access to this account [Y/n]? ');
+
+        if (['YES', 'Yes', 'yes', 'Y', 'y'].includes(answer)) {
+            const result = await account.deleteKey(options.accessKey);
+            inspectResponse.prettyPrintResponse(result, options);
+        } else {
+            rl.close();
+            console.log('Deleting key canceled.');
+        }
+    }
```
itegulov, line 43: We actually have a ready-to-use utility [function](https://github.com/near/near-cli/blob/master/utils/readline.js#L3) for asking y/n questions; you can use [this example](https://github.com/near/near-cli/blob/master/utils/check-credentials.js#L17-L21) as a reference.

TarasenkoFirelabs: Used `askYesNoQuestion`

TarasenkoFirelabs: Seems to me now in brach `master` we have this version of code:
   >>>>>>> master
    
    if(fullAccessKeys.length === 1 && fullAccessKeys[0].public_key.includes(options.accessKey)){
        console.log('You can\'t delete last FullAccess key');
    } else {
        const result = await account.deleteKey(options.accessKey);
        inspectResponse.prettyPrintResponse(result, options);
    }
    
    @volovyk-s 
should we close this PR cause we really not needed to ask customers are they really want to remove last fullAccess key - seems to me we already have proper decision in mater ?

commands/delete-key.js
```diff
@@ -20,6 +22,19 @@ async function deleteAccessKey(options) {
...
+            chalk`{bold.white You want to remove the last full access key and forgot access to this account ? {bold.green (y/n)}}`,
+            false
+        );
+        if (answer) {
+            const result = await account.deleteKey(options.accessKey);
+            inspectResponse.prettyPrintResponse(result, options);
+        } else {
+            console.log('Deleting key canceled.');
+        }
+    }
```
itegulov, line 39: Hmm, this means that the key will **only ever** get deleted if the account only has one full access key (even if the to-be-deleted key is of a completely different type). Let's follow @volovyk-s's [advice](https://github.com/near/near-cli/issues/834#issuecomment-968970448) here and make the flow look something like this:

```suggestion
    if (await checkIfSafeToDelete(account, options.accessKey)) {
        const result = await account.deleteKey(options.accessKey);
        inspectResponse.prettyPrintResponse(result, options);
    }
```

where `checkIfSafeToDelete` is a function that returns true if either the key is not a full access key or if the user agreed to delete it (twice if it is the last full access key). Something like this:
```js
async function checkIfSafeToDelete(account, accessKey) {
    let accessKeys = await account.getAccessKeys();
    let fullAccessKeys = accessKeys.filter(accessKey => accessKey.access_key.permission === 'FullAccess');

    if (fullAccessKeys.find(key => key.public_key === accessKey)) {
        const answer = await askYesNoQuestion(
            chalk`{bold.white Key [ {bold.blue ${accessKey}} ] is a full access key. Make sure it's not your recovery method. Do you want to proceed? {bold.green (y/n) }}`,
            false
        );
        if (!answer) {
            return false;
        }
        if (fullAccessKeys.length === 1) {
            const answer = await askYesNoQuestion(
                chalk`{bold.white Key [ {bold.blue ${accessKey}} ] is the last full access key. In case of deletion, you will not be able to restore access to the account [ {bold.blue ${accountId}} ]. Do you want to proceed? {bold.green (y/n) }}`,
                false
            );
            if (!answer) {
                return false;
            }
        }
    }
    return true;
}
```

commands/delete-key.js
itegulov: I am not sure where you found this code, but I can't see it anywhere in `master`. Do you have a link?

volovyk-s:
@TarasenkoFirelabs thank you for this contribution! I want to release it ASAP, so I moved it to [this](https://github.com/near/near-cli/pull/890) PR. Let's continue the discussion there.

PR #878, opened 2021-11-18
API keys support (MVP)

volovyk-s:
- Added API keys support (with config in file)
- Added short documentation in README file (a good place to start the review)


README.md
```diff
@@ -96,6 +96,19 @@ export NEAR_ENV=mainnet
 
 ---
 
+### API keys
```
frol, line 99: I want to mention that having "API keys" and "Access keys" might be confusing (in the rendered markdown file it is hard to notice that they have different header levels.

I would consider renaming "API keys" to "RPC server selection"

frol (approved):

README.md
volovyk-s: Ok, good point. Renamed to `Custom RPC server selection`, since we still have a `standard` one.

bin/near-cli.js
```diff
@@ -160,6 +160,11 @@ yargs // eslint-disable-line
         type: 'string',
         default: config.nodeUrl
     })
+    .option('headers', {
```
MaximusHaximus, line 163: IMO, we don't need to allow users to define arbitrary header contents at the CLI level for an initial release of custom RPC URL and API key support; :hocho: this entirely until we need it IMHO

MaximusHaximus (changes requested):
Added a couple of questions in-line <3

bin/near-cli.js
volovyk-s: This option is hidden and won't be shown to the users.

README.md
```diff
@@ -96,6 +96,18 @@ export NEAR_ENV=mainnet
 
 ---
 
+### Custom RPC server selection
+You can set your RPC server URL with the next command. Note, that this command will replace the RPC server for all networks. Clear it in case you want to get back to the default RPC server.
+```bash
+export NEAR_CLI_RPC_SERVER_URL=<put_your_rpc_server_url_here>
```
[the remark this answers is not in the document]
volovyk-s: @MaximusHaximus  Yes, I have not used exiting `NEAR_NODE _URL` because it's a breaking change and this design is only a POC. Also, these two env variables are used in two different scenarios. 
I just changed it for `localnet` to be:
```js
nodeUrl: process.env.NEAR_NODE_URL || 'http://localhost:3030',
```
Here is my reasoning.
- `NEAR_CLI_RPC_SERVER_URL` is something that you want to set and forget. This is for the daily usage.
- `NEAR_NODE_URL` is for development, validation, etc. Once you switched to `mainnet` or `testnet` you don't want `NEAR_NODE_URL` to affect your setup.

This solution is also not perfect, since there is no point to have just one variable for all networks. `Mainnet` RPC endpoint will not work for `testnet` and vice-versa. What do you think about a variable for each config?
`NEAR_CLI_MAINNET_RPC_SERVER_URL`
`NEAR_CLI_TESTNET_RPC_SERVER_URL`
`NEAR_CLI_LOCALNET_RPC_SERVER_URL` (+ `NEAR_NODE_URL` for backwards compatibility).
cc @frol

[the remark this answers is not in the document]
frol: I am fine with anything that works for you folks. I am a huge fun of explicit and stateless UX, so abusing env variables for network selection and configuration is already a lost battle to me :smile:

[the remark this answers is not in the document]
volovyk-s: Closing this discussion since the design is completely different now. Feel free to comment on the new code.

README.md
```diff
@@ -100,6 +100,34 @@ export NEAR_ENV=mainnet
...
 ---
 
+### Custom RPC server selection
+You can set custom RPC server URL by setting this env variables:
+```bash
+NEAR_CLI_MAINNET_RPC_SERVER_URL
+NEAR_CLI_TESTNET_RPC_SERVER_URL
+NEAR_CLI_BETANET_RPC_SERVER_URL
+NEAR_CLI_GUILDNET_RPC_SERVER_URL
+NEAR_CLI_GUILDNET_RPC_SERVER_URL
```
itegulov, line 110: `NEAR_CLI_GUILDNET_RPC_SERVER_URL` is repeated twice

config.js
```diff
@@ -1,72 +1,84 @@
...
             helperAccount: 'guildnet',
         };
+        break;
     case 'local':
     case 'localnet':
-        return {
+        config = {
             networkId: process.env.NEAR_CLI_LOCALNET_NETWORK_ID || 'local',
-            nodeUrl: process.env.NEAR_NODE_URL || 'http://localhost:3030',
+            nodeUrl: process.env.NEAR_CLI_GUILDNET_RPC_SERVER_URL || process.env.NEAR_NODE_URL || 'http://localhost:3030',
```
itegulov, line 57: Did you actually mean `NEAR_CLI_LOCALNET_RPC_SERVER_URL`?

README.md
volovyk-s: Thank you, fixed!

config.js
volovyk-s: The same :) Thank you!

itegulov (approved):

PR #883, opened 2021-11-23
Add a prompt warning about an existing contract on this account

### LINKED ISSUE #876: Add prompt to `near deploy`
Add prompt like:
`This account already has a deployed contract. Do you want to proceed?`
`-f`/`--force` flags should prevent such prompts.

itegulov:
Closes #876

This PR makes CLI output a prompt in the following format whenever an account already has an existing contract:
```
$ ./bin/near deploy --wasmFile ../NFT/res/non_fungible_token.wasm --accountId daniyar-near.testnet
Starting deployment. Account id: daniyar-near.testnet, node: https://rpc.testnet.near.org, helper: https://helper.testnet.near.org, file: ../NFT/res/non_fungible_token.wasm
This account already has a deployed contract [ 73ypSp7Ctk4eKh5odBbAD4H6jvYayKV7S2hhjTuMhp5R ]. Do you want to proceed? (y/n) n
```

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -122,12 +122,6 @@ const deploy = {
         })
         .alias({
             'accountId': ['account_id', 'contractName', 'contract_name'],
-        })
-        .option('force', {
-            desc: 'Forcefully deploy the contract even if there is already an existing contract',
-            type: 'boolean',
-            default: false,
-            alias: 'f'
         }),
     handler: exitOnError(main.deploy)
 };
@@ -225,6 +219,12 @@ yargs // eslint-disable-line
         type: 'boolean',
         alias: 'v',
         default: false
+    })
+    .option('force', {
+        desc: 'Forcefully execute the desired action even if it is unsafe to do so',
+        type: 'boolean',
+        default: false,
+        alias: 'f'
     })
     .middleware(require('../middleware/initial-balance'))
     .middleware(require('../middleware/print-options'))
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -122,6 +122,12 @@ const deploy = {
         })
         .alias({
             'accountId': ['account_id', 'contractName', 'contract_name'],
+        })
+        .option('force', {
+            desc: 'Forcefully deploy the contract even if there is already an existing contract',
+            type: 'boolean',
+            default: false,
+            alias: 'f'
         }),
     handler: exitOnError(main.deploy)
 };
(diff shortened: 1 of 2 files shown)


bin/near-cli.js
```diff
@@ -122,6 +122,12 @@ const deploy = {
         })
         .alias({
             'accountId': ['account_id', 'contractName', 'contract_name'],
+        })
+        .option('force', {
```
volovyk-s, line 126: Most likely this flag will be used to do `force` actions on other commands too (`delete-key` for example). I would make it more general.

index.js
```diff
@@ -39,6 +40,19 @@ exports.deploy = async function (options) {
     const account = await near.account(options.accountId);
     let prevState = await account.state();
     let prevCodeHash = prevState.code_hash;
+
```
volovyk-s, line 43: I would move `Starting deployment...` console.log to the place where it's actually happening now.

index.js
```diff
@@ -39,6 +40,19 @@ exports.deploy = async function (options) {
...
+    // Checks if there is an existing contract deployed on this account
+    // (code hash consisting of 32 ones means that the contract code is missing)
+    if (!options.force && prevCodeHash !== '11111111111111111111111111111111') {
+        const answer = await askYesNoQuestion(
+            chalk`{bold.white This account already has a deployed contract [ {bold.blue ${prevCodeHash}} ]. Do you want to proceed? {bold.green (y/n) }}`,
+            false
+        );
+        if (!answer) {
+            process.exit(1);
```
volovyk-s, line 52: We are trying not to terminate the process with `process.exit(x)`. I would suggest restructuring if/else checks.

volovyk-s (commented):
Looks good overall, left a few comments/suggestions :)

index.js
itegulov: Just for the sake of my curiosity can you elaborate on the reasoning here? I saw this pattern being used in some other places in the codebase so that is why I though it was okay to do so.

volovyk-s: `near-cli` is used for scripting, so I'm trying not to mess up with error codes.

volovyk-s: So, now `near deploy ...` -> `question` -> `No` will return 0 (success). And it's the right move since nothing is `failed`.
Not super important since people will not use this interactive mode in scripting, but that is a general rule for me.

volovyk-s (approved):
Thanks! Works just fine.

index.js
itegulov: Makes sense, thanks for the explanation.

PR #885, opened 2021-11-23
Added the ability to optionally set the keyPath and networkId for localnet

BenKurrek:
Fixes https://github.com/near/near-cli/issues/884

I've added the option to set the keyPath and networkId for localnet.

### DIFF
--- a/config.js
+++ b/config.js
@@ -47,9 +47,9 @@ function getConfig(env) {
     case 'local':
     case 'localnet':
         return {
-            networkId: 'local',
+            networkId: process.env.NETWORK_ID || 'local',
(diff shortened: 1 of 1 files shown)


volovyk-s (commented):
@BenKurrek LGTM, but let's keep naming more explicit.
`NEAR_CLI_LOCALNET_NETWORK_ID` and `NEAR_CLI_LOCALNET_KEY_PATH` will be more explanatory + we will not have problems in the future if we will decide to add such constants to other networks.

volovyk-s (approved):

volovyk-s:
Thank you for the contribution @BenKurrek !

PR #897, opened 2021-12-08
New seat price formula used

volovyk-s:
I will use npm dependency after the release of naj.

### DIFF
--- a/utils/validators-info.js
+++ b/utils/validators-info.js
@@ -8,15 +8,21 @@ async function validatorsInfo(near, blockNumberOrHash) {
         blockNumberOrHash = Number(blockNumberOrHash);
     }
     const genesisConfig = await near.connection.provider.sendJsonRpc('EXPERIMENTAL_genesis_config', {});
+    const protocolConfig = await near.connection.provider.sendJsonRpc('EXPERIMENTAL_protocol_config', { "finality": "final" });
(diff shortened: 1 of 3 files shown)


nikurt (approved):

utils/validators-info.js
```diff
@@ -8,15 +8,21 @@ async function validatorsInfo(near, blockNumberOrHash) {
...
     return result;
 }
 
 async function showValidatorsTable(near, blockNumberOrHash) {
     const result = await validatorsInfo(near, blockNumberOrHash);
-    const seatPrice = validators.findSeatPrice(result.current_validators, result.numSeats);
+    const seatPrice = validators.findSeatPrice(
+        result.current_validators,
+        result.numSeats,
+        result.genesisConfig.minimum_stake_ratio[1],
```
frol, line 24: Well, it is better to pass the actual ratio there, so if there is [2, 5], you would get 2/5, instead of 1/5. I feel we should fix it on near-api-js side.

frol (approved):
Other than the concern below, this LGTM!

utils/validators-info.js
volovyk-s: @frol created a PR: https://github.com/near/near-api-js/pull/746

frol (approved):

nikurt (approved):
Seat price calculation looks good.

PR #898, opened 2021-12-09
Number of seats changed to 1 in PV 49+

volovyk-s:
It supports both old and new Protocol Versions. I decided not to delete the column, because it's a breaking change.


nikurt (approved):
Tried the change locally, looks good.

I suggest to drop the "#Seats" column after the 1.23.0 release, as it no longer provides any additional value.

volovyk-s:
@nikurt I will drop it on the next breaking change and major version upgrade.

PR #913, opened 2022-01-13
Fix for https://github.com/near/near-cli/issues/786

june07:
If the seed phrase is used on the command line regardless of position, the seed-phrase middleware creates a key and adds it to the keystore.  This change considers that fact and eliminates output of the confusing message.

Also added the seedPhrase yargs option as it seems to be supported per how things work and even per the comment here:
https://github.com/near/near-cli/blob/63cd30cc27a3cefd9ff48c6d4e9b4b26816bb0ed/middleware/seed-phrase.js#L8

### DIFF
--- a/commands/generate-key.js
+++ b/commands/generate-key.js
@@ -47,7 +47,6 @@ module.exports = {
 
         // If key doesn't exist, create one and store in the keyStore.
         // Otherwise, it's expected that both key and accountId are already provided in arguments.
-
         if (!argv.publicKey) {
             const keyPair = KeyPair.fromRandom('ed25519');
             argv.publicKey = keyPair.publicKey.toString();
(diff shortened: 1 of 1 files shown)


june07:
🎤 ✔️ 12...

june07:
Any feedback on this one?

willemneal:
@volovyk-s

commands/generate-key.js
```diff
@@ -9,6 +9,10 @@ module.exports = {
         .option('yolo', {
             description: 'Do not ask for extra confirmation when using Ledger',
             type: 'boolean',
+        })
```
volovyk-s, line 12: This option is already available on the top level. We should not add it here again.

commands/generate-key.js
```diff
@@ -36,7 +40,7 @@ module.exports = {
 
         const { deps: { keyStore } } = near.config;
         const existingKey = await keyStore.getKey(argv.networkId, argv.accountId);
-        if (existingKey) {
+        if (existingKey && !argv.seedPhrase) {
```
volovyk-s, line 43: It will fix just 1 out of ~5 problems we have with generation keys. This is a critical please of code, so I would prefer to fix them all at once.

PR #926, opened 2022-02-03
Fix account undefined not found

### LINKED ISSUE #911: Fix: Account undefined is not found…
It looks like there's a small issue as shown in this screenshot:

Screen Shot 2022-01-04 at 3 28 29 PM

You can reproduce by running:

    near view pretend.mike.testnet fake_method ''

It should not say "undefined" and instead say the name of the NEAR account.

narwalsandeep:
Fixes #911

### DIFF
--- a/utils/exit-on-error.js
+++ b/utils/exit-on-error.js
@@ -32,7 +32,6 @@ process.on('exit', () => {
 });
 
 module.exports = (promiseFn) => async (...args) => {
-
     const command = args[0]['_'];
     process.env.NEAR_CLI_ERROR_LAST_COMMAND = command;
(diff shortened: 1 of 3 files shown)


narwalsandeep:
@mikedotexe PR

utils/exit-on-error.js
```diff
@@ -32,6 +32,7 @@ process.on('exit', () => {
 });
 
 module.exports = (promiseFn) => async (...args) => {
+
```
willemneal, line 35:
```suggestion
```

willemneal: Thanks! I have no power here and @volovyk-s is the better bet when tagging for a review.

narwalsandeep:
@volovyk-s plz review

PR #931, opened 2022-02-10
feat(dev-deploy): added ability to specify custom folder for generati…

chefsale:
…ng the keys


willemneal (approved):

volovyk-s:
@chefsale reverted. Can not recreate this PR for some reason.
The code looks good, but let's get rid of changes in a lock file.
Version is updated during the release process, so that change needs to be deleted too.
Please, create a new PR and I will release the new version shortly.

PR #938, opened 2022-02-15
added new arg base64file to support loading base64 blob from file int…

### LINKED ISSUE #937: Support base64 encoded files as arguments (fixes linux ulimit issues)
**Is your feature request related to a problem? Please describe.**
On linux, submitting long args list using -base64 flag can actually require more bytes than are allowed by system. It is not possible to sign transactions without a different approach (similar to deploying wasm)

**Describe the solution you'd like**
See the following line:
* https://github.com/near/near-cli/blob/master/commands/call.js#L54

This could get upgraded to using a similar file read pattern like when deploying:
```javascript
const actions = [transactions.deployContract(readFileSync(options.wasmFile))];
```
So it could be like: 
```javascript
const base64Args = readFileSync(options.base64File);
```

**Describe alternatives you've considered**
I have tried to utilize xargs & change the file limits, however it will not work for larger data. (im storing bytes of a smart contract for doing upgrades via DAOs)

**Additional context**
You can see some CLI commands we're testing: 
https://github.com/near-daos/sputnik-dao-contract/pull/111

TrevorJTClarke:
…o CLI args

This PR fixes #937 
Directly impacting any CLI args on linux

### DIFF
--- a/commands/call.js
+++ b/commands/call.js
@@ -1,3 +1,4 @@
+const { readFileSync } = require('fs');
 const { DEFAULT_FUNCTION_CALL_GAS, providers, utils } = require('near-api-js');
 const exitOnError = require('../utils/exit-on-error');
 const connect = require('../utils/connect');
@@ -39,6 +40,11 @@ module.exports = {
             required: true,
             desc: 'Unique identifier for the account that will be used to sign this call',
             type: 'string'
+        })
+        .option('base64file', {
+            desc: 'Load a base64-encoded BLOB file into the args sent to a function call.',
+            type: 'string',
+            default: false
         }),
     handler: exitOnError(scheduleFunctionCall)
 };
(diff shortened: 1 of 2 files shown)


commands/call.js
```diff
@@ -51,7 +57,19 @@ async function scheduleFunctionCall(options) {
...
-    const parsedArgs = options.base64 ? Buffer.from(options.args, 'base64') : JSON.parse(options.args || '{}');
+    let parsedArgs;
+    // load args rare order first
+    if (options.base64file) {
+        try {
+            let fileargs = await readFileSync(options.base64file);
+            parsedArgs = Buffer.from(fileargs, 'base64');
+            console.log(`Loaded base64 args file, size ${parsedArgs.length}`);
+        } catch (e) {
+            console.log('Could not load base64 file!');
```
willemneal, line 68: Shouldn't this throw so as exit?  Currently all errors throw and you end up with a large noisy help message, but there are plans to refactor how errors are handled.

TrevorJTClarke:
@willemneal looks like the last build failed because RPC 502: https://app.travis-ci.com/github/near/near-cli/builds/246520845#L260

commands/call.js
TrevorJTClarke: Good call, changed!

TrevorJTClarke:
> @willemneal looks like the last build failed because RPC 502: https://app.travis-ci.com/github/near/near-cli/builds/246520845#L260

@willemneal should i trigger with a commit, or can you manually trigger a rebuild? I think RPC services shipped recently are fixed now.

willemneal:
@TrevorJTClarke I wish I could, but I'm not a collaborator.  So you could maybe do a push again or a small change and push, then revert and force push?

TrevorJTClarke:
@willemneal nevermind, same RPC 502... Some config or something got changed. No idea how to help get this across the line. Super needed in timely manner - Auditors starting in 8 days need this

willemneal:
@TrevorJTClarke You can depend on your version of near-cli.  `npm install ...`

I had to do that while waiting for this PR: https://github.com/near/near-cli/pull/903, which has been released and could also be used to solve your issue.

gagdiez:
This has been stalled for more than 2 years, I will proceed to close it and we can re-open it if the community requires it.

With that being said, I think it is quite simple to do this in the command line, we actually show how to do it when relevant in our docs: https://docs.near.org/tutorials/examples/factory#update-the-stored-contract

PR #944, opened 2022-02-24
943 initFunction and initArgs logic is wrong

### LINKED ISSUE #943: initFunction and initArgs logic is wrong
**Describe the bug**
When user specifies `initArgs` but does not specify `initFunction` the error message intended by the code here does not show: https://github.com/near/near-cli/blob/master/commands/dev-deploy.js#L87-L98 . It appears the logic in this code is wrong.

**To Reproduce**
1. Deploy and initialize the contract using `initArgs` but omitting `initFunction`, e.g.:
```bash
near deploy --accountId=mycontract01.testnet --wasmFile=res/ft_lockup.wasm --initArgs '{<add args here>}'
```
2. Observe that no error message is created and initialization function is not called.

**Expected behavior**
Error message is displayed OR `new` function is used by the default.

willemneal:
Use "new" if `--initArgs` is provided but `--initFunction` is not. fixes #943 

@nearmax 
@volovyk-s

### DIFF
--- a/bin/near-cli.js
+++ b/bin/near-cli.js
@@ -105,7 +105,8 @@ const deploy = {
         })
         .option('initFunction', {
             desc: 'Initialization method',
-            type: 'string'
+            type: 'string',
+            default: 'new'
         })
         .option('initArgs', {
             desc: 'Initialization arguments',
--- a/commands/dev-deploy.js
+++ b/commands/dev-deploy.js
@@ -27,7 +27,8 @@ module.exports = {
         })
         .option('initFunction', {
             desc: 'Initialization method',
-            type: 'string'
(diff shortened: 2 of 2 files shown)


nearmax (changes requested):
The entire logic can be fixed by setting `default` of `initFunction` to `"new"`.

volovyk-s (approved):

kalloc:
Hi, but what about contract re-deploy/upgrade? Is I understand that  initFunction for migration not always required.

volovyk-s:
@kalloc yes, `new` as default caused some bugs. Addressed in https://github.com/near/near-cli/pull/969

PR #962, opened 2022-04-14
Mixpanel location

volovyk-s:
image

### DIFF
--- a/utils/eventtracking.js
+++ b/utils/eventtracking.js
@@ -53,7 +53,7 @@ const track = async (eventType, eventProperties, options) => {
             const id = getMixpanelID(shellSettings);
             await Promise.all([
                 mixpanel.alias(accountID, id),
-                mixpanel.people.set(id, {account_id: accountID})
+                mixpanel.people.set(id, { account_id: accountID })
             ]);
         }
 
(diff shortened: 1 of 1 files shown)


utils/eventtracking.js
```diff
@@ -82,6 +86,16 @@ const track = async (eventType, eventProperties, options) => {
...
+    ).then(
+        res => res.text()
+    ).then(
+        ip => ip.trim()
+    );
+}
+
 const getEventTrackingConsent = async () => {
     return askYesNoQuestion(
         chalk`Please help us to collect data on near-cli usage to improve developer experience. ` +
```
TiffanyGYJ, line 101: Is this the only question for getting consent? Do we have separate one when we grab their account id?

TiffanyGYJ (approved):

utils/eventtracking.js
volovyk-s: This is the only question that is getting asked on the first run.
We are not collecting NEAR Account ID, we are collecting generated user ID. It's unique for each machine and can be easily deleted by the user.

TiffanyGYJ:
Any reason why the checks are failed? Is that why the change has not been merged?

volovyk-s:
@TiffanyGYJ Travis is dead for some reason. Trying to figure out the reason. I will release it right after that.

PR #969, opened 2022-04-25
GitHub actions added

volovyk-s:


commands/dev-deploy.js
```diff
@@ -85,6 +84,10 @@ async function devDeploy(options) {
     // Deploy with init function and args
     const actions = [transactions.deployContract(readFileSync(options.wasmFile))];
 
+    if (options.initArgs && !options.initFunction) {
+        options.initFunction = 'new';
```
ailisp, line 88: Why we use this over `default: 'new'` ?

ailisp (approved):
CI setup looks good. The eslint danger is failing, not sure if that need to be fixed

volovyk-s:
@ailisp I will fix danger in a separate PR. Or just set API keys, looks like this is the problem.

commands/dev-deploy.js
volovyk-s: @ailisp The idea is to be able not to set `initFunction` in CLI when it has the default name (new).
But when we are setting the default value of `initFunction` to new, args become mandatory even when you are not initializing the contract and it breaks the logic.

PR #975, opened 2022-05-17
feat: js command

ChaoticTempest:
Initial work on the JS cli command which includes:
- deploy
- call
- remove

Getting this one out for now to see if we want to move in this kind of direction before going on to adding more subcommands


volovyk-s:
@ChaoticTempest Looks cool. Let's fix the bug and merge.

commands/js.js
```diff
@@ -0,0 +1,229 @@
...
+// Encode JSVM related arguments into base64 for use in the enclave
+function base64_encode_args(contractId, functionName, args) {
+    let buf = Buffer.concat([
+        Buffer.from(contractId),
+        Buffer.from([0]),
+        Buffer.from(functionName),
+        Buffer.from([0]),
+        Buffer.from(args)]
+    );
+    return Buffer.from(buf.toString('base64'), 'base64');
```
ailisp, line 126: This is a invariant operation. 
```
> Buffer.from([0])
<Buffer 00>
> Buffer.from([0]).toString('base64')
'AA=='
> Buffer.from(Buffer.from([0]).toString('base64'), 'base64')
<Buffer 00>

```


it's equivalent to simply return Buffer:
```suggestion
    return buf
```

ailisp (approved):
Great job! It works, we may want some small enhancement after this PR.

ailisp:
@ChaoticTempest Looks like the deploy_js_contract ignore the initFunction. As we don't have the support to batch deploy and call init function in the same transaction, there're two choices here:
1. Drop `initFunction` option, let user call a `near js call contract init ...` right after the deploy
2. Keep `initFunction`, in deploy cmd, do deploy and call as two transactions

ChaoticTempest:
@ailisp I don't think either of those options would guarantee us the safety of a batch transaction though unless point 2 is talking about that. JSVM should be able to schedule a promise that would allow us to do this `deploy` with the `initFunction`. But I get that it could take a while to implement and since our aim is June to release, we likely won't make it in time. I'll remove the init* related args and we can add it back in later

ailisp:
It looks very solid and all good on my side! But I'm not often in maintaining near-cli, so let @volovyk-s double check it fits near-cli well.

volovyk-s:
@ChaoticTempest we have working tests in master now. I will add them here.

volovyk-s:
Nice work @ChaoticTempest ! Merging now. Let's add some docs and examples from your doc to the  README in the next PR.

commands/js.js
```diff
@@ -0,0 +1,324 @@
...
+        .command(js_call)
+        .command(js_view)
+    ,
+};
+
+// Encode JSVM related arguments into base64 for use in the enclave
+function base64_encode_args(contractId, functionName, args) {
+    let buf = Buffer.concat([
+        Buffer.from(contractId),
+        Buffer.from([0]),
```
austinabell, line 184: curiosity: why are 0 bytes included in-between these?

PR #981, opened 2022-05-20
Danger fixed

volovyk-s:
`danger-plugin-yarn` is broken. The whole point of Danger setup was the prevention of `.lock` file modification without modification of `package.json`, so I have done just that.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,4 +1,4 @@
-const { danger, fail, warn } = require('danger');
+const { danger, warn } = require('danger');
 const { includes } = require('lodash');
 
 const hasPackageChanges = includes(danger.git.modified_files, 'package.json');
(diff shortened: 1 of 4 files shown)


dangerfile.js
```diff
@@ -1,15 +1,8 @@
...
-    checkForNewDependencies(packageDiff);
-    checkForLockfileDiff(packageDiff);
-    checkForTypesInDeps(packageDiff);
-});
\ No newline at end of file
+const hasPackageChanges = includes(danger.git.modified_files, 'package.json');
+const hasLockfileChanges = includes(danger.git.modified_files, 'yarn.lock');
+if (hasPackageChanges && !hasLockfileChanges) {
+    warn('There are package.json changes with no corresponding lockfile changes');
+}
```
ChaoticTempest, line 8: do we want to warn about this case as well?
```js
if (!hasPackageChanges && hasLockfileChanges)
```

volovyk-s: @ChaoticTempest No, because package.json can have changes that are not related to dependencies.

ChaoticTempest (approved):

PR #982, opened 2022-05-20
User country info

volovyk-s:


utils/eventtracking.js
```diff
@@ -53,15 +53,18 @@ const track = async (eventType, eventProperties, options) => {
             const id = getMixpanelID(shellSettings);
             await Promise.all([
                 mixpanel.alias(accountID, id),
-                mixpanel.people.set(id, {account_id: accountID})
+                mixpanel.people.set(id, { account_id: accountID })
             ]);
         }
 
+        const user_country = await getUserCountry();
```
ChaoticTempest, line 60: when `getUserCountry` throws an error and `console.log()` it, won't the return value be `undefined` here? And in that case, shouldn't we be gracefully handling it here or in the function itself?

utils/eventtracking.js
```diff
@@ -82,10 +85,18 @@ const track = async (eventType, eventProperties, options) => {
     }
 };
 
+async function getUserCountry() {
+    return fetch('https://ipinfo.io/json').then(
+        (response) => response.json()
+    ).then(
+        (jsonResponse) => jsonResponse.country
+    ).catch(error => console.log(error));
```
ChaoticTempest, line 93: Any difference if we use `console.error` over `console.log` here?

utils/eventtracking.js
volovyk-s: Now we will just log the message. How do you want to handle it? If this call was unsuccessful - we should not return any errors. The user does not care if it was successful. Same reason for `.log()` vs `.error()`. Also, it's ok to send `undefined`, because this is what happened. The country is undefined.

utils/eventtracking.js
volovyk-s: Addressed above.

utils/eventtracking.js
ChaoticTempest: I see, if undefined is expected then its fine

utils/eventtracking.js
```diff
@@ -82,10 +85,18 @@ const track = async (eventType, eventProperties, options) => {
     }
 };
 
+async function getUserCountry() {
+    return fetch('https://ipinfo.io/json').then(
+        (response) => response.json()
+    ).then(
+        (jsonResponse) => jsonResponse.country
+    ).catch(() => console.log('Failed to get the country'));
```
ChaoticTempest, line 93: how about we also format the error just so we don't silently ignore it:
```suggestion
    ).catch(error => console.log(`Failed to get the country due to: ${error}`));
```

ChaoticTempest (approved):

PR #983, opened 2022-05-20
chore: Remove requirement for specifying accountId for js view

ChaoticTempest:
this forces people to specify the signer for `near js view`, so removed the requirement and have the signer be whatever. This isn't really even signing anything, just an oddity of `near-api-js` requiring an account-id to create an account, to then make a view call

### DIFF
--- a/commands/js.js
+++ b/commands/js.js
@@ -150,7 +150,6 @@ const js_view = {
         .option('accountId', {
             desc: 'Unique identifier for the account that will be used to sign this call',
             type: 'string',
-            required: true,
         })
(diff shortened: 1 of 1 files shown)


volovyk-s:
Same situation as in https://github.com/near/near-cli/pull/984

ChaoticTempest:
this one's a little different, since this is due to NAJ requiring an accountId/account object to send a view function call. Specifying the accountId here isn't needed at all in the CLI so making it optional to not create a breaking change

BenKurrek (approved):

PR #984, opened 2022-05-20
feat: Make --accountId optional for js call

ChaoticTempest:
Optional convenience change for `near js call`. This will not require the user to specify the `--accountId` option if they wish to use the same id as the `contractId`.

So before, we'd have:
```sh
near js call contract-id function-name --args <args> --accountId contract-id
```
Now, we can just do:
```sh
near js call contract-id function-name --args <args>
```

### DIFF
--- a/commands/js.js
+++ b/commands/js.js
@@ -111,9 +111,9 @@ const js_call = {
     desc: 'Call a method on a contract',
     builder: (yargs) => yargs
         .option('accountId', {
-            desc: 'Unique identifier for the account that will be used to sign this call',
+            desc: 'Unique identifier for the account that will be used to sign this call. Will use the contract-id if not specified',
             type: 'string',
-            required: true,
+            default: null,
         })
         .option('args', {
             desc: 'Arguments to the contract call, in JSON format by default (e.g. \'{"param_a": "value"}\')',
@@ -290,6 +290,10 @@ async function remove(options) {
 }
 
 async function call(options) {
+    if (options.accountId === null) {
+        options.accountId = options.contractId;
+    }
+
     await checkCredentials(options.accountId, options.networkId, options.keyStore);
 
     const jsvmId = jsvm_contract_id(options);


commands/js.js
```diff
@@ -111,9 +111,9 @@ const js_call = {
     desc: 'Call a method on a contract',
     builder: (yargs) => yargs
         .option('accountId', {
-            desc: 'Unique identifier for the account that will be used to sign this call',
+            desc: 'Unique identifier for the account that will be used to sign this call. Will use the contract-id if not specified',
             type: 'string',
-            required: true,
+            default: null,
         })
```
ChaoticTempest, line 117: @volovyk-s is there some clever/cleaner way to do what I'm doing with setting the `default: null`, but maybe tell yargs that we want the value of `options.contractId`?

volovyk-s: @ChaoticTempest if `accountId` is not necessary - we should delete this option.

Is it possible that `accountId` can affect the result somehow?

`accountId` option deletion is a breaking change. It means that we should keep it, but log a warning that it's deprecated. `default: 'string'` line should be deleted and `accountId` should not be used anywhere.

ChaoticTempest: accountId is still required in some cases here since user might want to specify a different signer than the contractId. This is just a convenience addition so that people don't need to specify it twice, so no breaking change

volovyk-s: Oh, I got it wrong. I was sure that we are talking about a view call.

commands/js.js
```diff
@@ -111,9 +111,9 @@ const js_call = {
     desc: 'Call a method on a contract',
     builder: (yargs) => yargs
         .option('accountId', {
-            desc: 'Unique identifier for the account that will be used to sign this call',
+            desc: 'Unique identifier for the account that will be used to sign this call. Will use the contract-id if not specified',
             type: 'string',
-            required: true,
+            default: null,
```
volovyk-s, line 116:
```suggestion
```

commands/js.js
```diff
@@ -290,6 +290,10 @@ async function remove(options) {
 }
 
 async function call(options) {
+    if (options.accountId === null) {
```
volovyk-s, line 293:
```suggestion
    if (!options.accountId) {
```

commands/js.js
ChaoticTempest: hmm, yeah thinking about it again, let's just be explicit here and stay consistent with regular near-cli call command

PR #991, opened 2022-06-16
lower account creation initial balance from 100 to 10

### LINKED ISSUE #990: Reduce default amount of $NEAR to create an account from 100 to 10
People often get stuck in the account creation process with "not enough balance" and don't realize the default is 100. We should lower this default to 10 as during development users only get 200 $NEAR `testnet` tokens and often try to create multiple sub-accounts.

thisisjoshford:
Fixes #990

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,7 +488,7 @@ near create-account sub-acct.example-acct.testnet --masterAccount example-acct.t
 **Example using `--initialBalance`:**
 
 ```bash
-near create-account sub-acct2.example-acct.testnet --masterAccount example-acct.testnet --initialBalance 10
+near create-account sub-acct2.example-acct.testnet --masterAccount example-acct.testnet --initialBalance 1
 ```
 
 <details>
--- a/commands/create-account.js
+++ b/commands/create-account.js
@@ -34,7 +34,7 @@ const createAccountCommand = {
         .option('initialBalance', {
             desc: 'Number of tokens to transfer to newly created account',
             type: 'string',
-            default: '100'
+            default: '10'
         }),
     handler: exitOnError(createAccount)
 };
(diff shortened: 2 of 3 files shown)


volovyk-s:
It will be a breaking change for a lot of scripts that are relying on this amount. I would rather not merge it.
Isn't it better to increase the number of tokens in the original acc from 200 to ~1000?

willemneal:
I still don't understand why this couldn't be a major bump then.  I assume those are test scripts that could be updated to use the newer version.  From a devx perspective it's a weird default.

thisisjoshford:
> I still don't understand why this couldn't be a major bump then. I assume those are test scripts that could be updated to use the newer version. From a devx perspective it's a weird default.

Yeah, I do agree that 100 seems to be a weird default and yeah... most likely these scripts are tests that will need to be updated for users. Thoughts @volovyk-s ?

> Isn't it better to increase the number of tokens in the original acc from 200 to ~1000?
I think if we did this the faucets would get depleted very quickly as some write scripts to create a dev-account then delete it and grab the 200 $NEAR. We did have the default set to 500 $NEAR and then lowered it... I think it was because someone was draining testnet faucets but not sure how lowering it by 300 resolved the issue. Let me see if I can dig up the reasoning of lowering initial balance of dev-accounts / testnet accounts.

MaximusHaximus:
FYI, testnet account creation that doesn't use a `masterAccount` to fund the account (e.g. the `dev-deploy` command, which only works on testnet) would not be impacted by this change, because testnet account creation is provided _by a REST API call to the near-contract-helper_, which ignores any value you provide and _always_ gives every account the same 200N on creation.  

FYI, the NEAR that are given to new `dev-deploy` created accounts are provided from the balance held by the `testnet` top-level account, which has a balance currently of over 55 million NEAR (https://explorer.testnet.near.org/accounts/testnet).

It is not clear to me whether, how, and/or how often the `testnet` account gets additional balance added to it, but if it is never increased, then using 200N per account, we could create around ~275,000 dev-deploy accounts before exhausting the current balance of the `testnet` account.  Is that something you are investigating already @thisisjoshford ?

PR #997, opened 2022-06-23
Create seed phrase option on add-key

tituszban:
The problem:
There is currently no way in the CLI to add a seed phrase to an already existing account. As a result, if an account was generated on the CLI, it is not easy to also add it on the web wallet.

The solution:
If `add-key` is called not with a public key, but with the `--seedPhrase` option, the key generated by that seed phrase is added to the account

### DIFF
--- a/commands/add-key.js
+++ b/commands/add-key.js
@@ -5,13 +5,13 @@ const { utils } = require('near-api-js');
 const checkCredentials = require('../utils/check-credentials');
 
 module.exports = {
-    command: 'add-key <account-id> <access-key>',
+    command: 'add-key <account-id>',
     desc: 'Add an access key to given account',
     builder: (yargs) => yargs
         .option('access-key', {
             desc: 'Public key to add (base58 encoded)',
             type: 'string',
-            required: true,
+            required: false,
         })
         .option('contract-id', {
             desc: 'Limit access key to given contract (if not provided - will create full access key)',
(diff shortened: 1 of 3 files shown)


MattKetmo:
yes my bad, I confused add-key & generate-key when talking about "overwrites".

A few things about your code:

- Setting `access-key` `required` to `false` changes the behavior of the command. 
  It's no more a 2nd arg but an option `--access-key`. Which means part 1 in README should be edited too (as well as doc pages)

- Also, since it's a non-required option, a non-empty check should be added before calling `await account.addKey(options.accessKey` at line 55 in order to avoid this error

```
Adding full access key = undefined to example.testnet.
An error occured
TypeError: Cannot read properties of undefined (reading 'borshSerialize')
    at serializeStruct (/private/tmp/near-cli/node_modules/borsh/lib/index.js:302:20)
...
```

- I would not `console.log` the mnemonic for security reasons (or maybe you wanted to write `seedPhrasePublicKey` at line 44?)

Otherwise yes I think this improve ux. We could even have an option to auto generate the seed phrase.

tituszban:
* Updated the readme, and PR to update the docs: https://github.com/near/docs/pull/1102
* Added a non-empty check
* Removed seed phrase from console log

commands/add-key.js
```diff
@@ -5,13 +5,13 @@ const { utils } = require('near-api-js');
 const checkCredentials = require('../utils/check-credentials');
 
 module.exports = {
-    command: 'add-key <account-id> <access-key>',
+    command: 'add-key <account-id>',
```
MattKetmo, line 8: I noticed there is another syntax to possibly use the existing arg as an optional parameter
```suggestion
    command: 'add-key <account-id> [access-key]',
```

example in the [generate-key command](https://github.com/near/near-cli/blob/4b4a787d0eafaa97fb7c2c741cdfaacf2c53f35b/commands/generate-key.js#L6)

This would allow to not break the current input of the command (and thus the existing docs). But I let maintainers decide for the best solution :)

tituszban: I think that's a pretty good idea. Other thoughs?

PR #999, opened 2022-06-27
refactor: check if beneficiary account exists, require confirmation

judith-Near:
related to issue https://github.com/near/near-cli/issues/994

PR adds a warning `This method will delete your account. Beneficiary account must exist in order to transfer all Near tokens. Make sure to send all fungible tokens or NFTs that you own to the beneficiary account prior to deleting, as this method will only transfer NEAR tokens. Do you want to proceed?` that the user must confirm.

Delete will also not run if beneficiary account does not exist.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,7 @@ const confirmDelete = function () {
     return askYesNoQuestion(
         chalk`{bold.white This method will delete your account. Beneficiary account must exist in order to transfer all Near tokens. Make sure to send all fungible tokens or NFTs that you own to the beneficiary account prior to deleting, as this method will only transfer NEAR tokens. Do you want to proceed? {bold.green (y/n) }}`,
         false);
-}
+};
 
 exports.deploy = async function (options) {
     await checkCredentials(options.accountId, options.networkId, options.keyStore);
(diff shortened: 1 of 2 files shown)


BenKurrek (approved):
lgtm

volovyk-s:
@judith-Near @BenKurrek looks like this PR broke tests.

BenKurrek:
@volovyk-s oop yea it looks like the tests have to be rewritten?

index.js
```diff
@@ -215,13 +221,25 @@ exports.viewAccount = async function (options) {
...
-    console.log(
-        `Deleting account. Account id: ${options.accountId}, node: ${options.nodeUrl}, helper: ${options.helperUrl}, beneficiary: ${options.beneficiaryId}`);
     const near = await connect(options);
-    const account = await near.account(options.accountId);
-    const result = await account.deleteAccount(options.beneficiaryId);
-    inspectResponse.prettyPrintResponse(result, options);
-    console.log(`Account ${options.accountId} for network "${options.networkId}" was deleted.`);
+    const beneficiaryAccount = await near.account(options.beneficiaryId);
+    // beneficiary account does not exist if there are no access keys
+    if (!(await beneficiaryAccount.getAccessKeys()).length) {
```
volovyk-s, line 227: @judith-Near If the account has 0 AC. itdoes not mean it does not exist.

volovyk-s: + geAccessKeys() can return `null`, it will crash the execution.

BenKurrek: Ah... I misunderstood what's happening here. I was under the impression that this would throw an error if the account doesn't exist and you're trying to query for the keys. That's on me. Apologies.

volovyk-s: Hm, maybe it will. It's better to test it. But logic seems wrong anyway.

idea404: Addressed here: https://github.com/near/near-cli/pull/1009/files

PR #1003, opened 2022-07-20
Fixing generate-key with --seedPhrase

mm-near:
Currently when you call

near --generate-key --seedPhrase a b c .. 

It will create 2 implicit keys (one with the seedPhrase and the other one with random value) - which is a little bit confusing.

Moreover, I think that something's wrong with seed-phrase.js code - as it should probably store the key in a seedPhraseKeystore (but I'm not very familiar with this code).


volovyk-s:
@judith-Near can you please fix tests, since they are broken on master after your last merge.

PR #1005, opened 2022-07-22
feat: migrate from mixpanel to segment

itegulov:


utils/eventtracking.js
```diff
@@ -1,13 +1,15 @@
-const MIXPANEL_TOKEN = 'e98aa9d6d259d9d78f20cb05cb54f5cb';
+const SEGMENT_WRITE_KEY = 'vErt52x0024wPxUfLDZIKcYRPQgh6NyT';
```
itegulov, line 1: Using my personal key for testing, but I need an official NEAR key before this is merged. Not sure who owns the Segment account.

CC @AustinBaggio @TiffanyGYJ

utils/eventtracking.js
```diff
@@ -117,8 +127,8 @@ const askForConsentIfNeeded = async (options) => {
 
 const trackDeployedContract = async () => {
     const shellSettings = settings.getShellSettings();
-    const id = getMixpanelID(shellSettings);
-    await mixpanel.people.increment(id, 'deployed_contracts');
+    const id = getSegmentID(shellSettings);
+    // TODO: find a way to increment a Segment analytic property
```
itegulov, line 131: It seems like there is no persistent analytic properties in segment (or at least a direct replacement). Not sure how much we care about the total amount of deployed contracts as a stored number vs aggregating them by `deployed` events.

utils/eventtracking.js
itegulov:
```suggestion
const SEGMENT_WRITE_KEY = 'ePJM5UVxZGUxLP9dH2wGDvUE7hC2MbTX';
```

utils/eventtracking.js
```diff
@@ -48,19 +50,20 @@ const track = async (eventType, eventProperties, options) => {
             return;
         }
 
-        if (shouldTrackID(shellSettings)) {
+        if (options.accountId && shouldTrackID(shellSettings)) {
```
volovyk-s, line 53: @itegulov is this change intended?
I think users can avoid tracking even if they do not pass any `accountId` in command.

utils/eventtracking.js
volovyk-s, line 131: @itegulov are you planning to work on it now?
Deployed contracts are the most important part of analytics IMO.

volovyk-s (commented):
It's great that the API is so similar!

utils/eventtracking.js
volovyk-s: @TiffanyGYJ

itegulov: I have spent some time looking for alternatives to no avail, so I think it would make sense to hear @TiffanyGYJ's opinion first before spending even more time on this

itegulov: > Deployed contracts are the most important part of analytics IMO.

Curious to hear your reasoning. Wouldn't aggregating `deployed` events achieve the same goal? Do we need to store/cache this as a definite number for something?

TiffanyGYJ: Deployed events might happen multiple times with different contracts, so that might not be an accurate estimate on how many unique contracts each dev deployed. 
Is there a way we can track that somewhere else or finding a way from Segment/Mixpanel is the best option?

itegulov: @TiffanyGYJ 
> Deployed events might happen multiple times with different contracts, so that might not be an accurate estimate on how many unique contracts each dev deployed.

Looking through the code, the current logic does not track unique deployed contracts either, just the number of times when `devDeploy`/`deploy` were executed successfully.

volovyk-s: I think that the total number of deploys per week is the most important metric since we are tracking developer activity.
The number of active contracts (contracts that have been deployed or redeployed in the past week) is also important.
Tracking a particular user and his deploys is not that important IMO, and a bit shady.

itegulov: > I think that the total number of deploys per week is the most important metric since we are tracking developer activity. The number of active contracts (contracts that have been deployed or redeployed in the past week) is also important.

I agree with this, but a per-user counter does not really help with that, no? Increments do not have a timestamp, so you would have to track how much a counter has changed each week which kind of beats its point. I think we can deduce all that information just from tracking events (by writing a small custom event consumer or building something in Segment itself, not sure if it provides any processing logic).

> Tracking a particular user and his deploys is not that important IMO, and a bit shady.

:+1:

volovyk-s: Agree!

itegulov:
I have deleted the deployed contract tracking, so I think this PR is ready for a review

CC @volovyk-s @TiffanyGYJ

volovyk-s (approved):

Issue #796, opened 2021-06-15
feat: near dev deploy improve CI friendliness


zcstarr:
I'm adding near dev deploy as a ci step. The way near dev deploy is setup at the moment has an event tracking prompt that can't be avoided the first time you run it. This prompt can hang ci unless you redirect some input into stdin. This can get a bit hairy as the context changes.

[https://github.com/near/near-cli/blob/90ca53c8ffcd1fdac7b327fb5a6afcfdc6dc58b1/commands/dev-deploy.js#L42](https://github.com/near/near-cli/blob/90ca53c8ffcd1fdac7b327fb5a6afcfdc6dc58b1/commands/dev-deploy.js#L42)

How do people feel about bubbling up the consent option as an optional arg. I know this gets weird when you save the state in shell settings. The arg would be more of an override you use the arg vs use what's in shell settings?

maxhr:
Bumping this, as it breaks my CI too.

idea404:
Bump as I'd like to run a CI that tests tutorial commands and this is a blocker. I would propose changing this to a notification that does not require user input which defaults at "no" for sharing usage data and can be thus changed at a later time.

angshumanHalder:
Did you guys find any workaround to this? I was trying to dockerize my app and stuck in the deploy phase as the cli doesn't take any default input for tracking, the build process is failing.

idea404:
@angshumanHalder you can pass the 'y' command: `printf 'y\n'` | near dev-deploy

angshumanHalder:
@idea404 thanks. I am able to automate the deploy and build processes. 😄

PR #1010, opened 2022-08-08
Notify users about tracking opt-in instead of forcing choice

idea404:
This PR addresses the data-sharing opt-in that prompts user feedback in the current version of NEAR CLI. This opt-in prompt requires user input and complicates implementing NEAR CLI in CI pipelines in straightforward manner. (Further mentioned in #796)

The changes are:

- NEAR CLI now _notifies_ the user of the possibility of opting in for data sharing. The setting is defaulted to `false`. 
- User may opt in by running `near track yes` and opt out at any later time by running the same command with `no` as a parameter
- CI test is introduced to test this functionality
- These changes preserve previous tracking behavior that does **not** add an `accountId` to the data shared

Notification:
image

Enable tracking:
image

Disable tracking:
image

Wrong param:
image


volovyk-s:
I'm sure that the amount of data that we receive will drop significantly. Nobody will explicitly tell the system that they want to share usage statistics. Was this design discussed with the product teamt? cc @AustinBaggio @TiffanyGYJ

As an alternative, we can add the  `near tracking enable/disable` command but leave the prompt as it is.
We will still revive user statistics and people who write scripts will start them with `near tracking enable/disable`.

Another option - add `--yes` flag. Check how it is done in the `npm init` command.

TiffanyGYJ:
This is a good point, thank you for bringing it up.
For the near tracking enable/disable, will we auto-enroll them into enable or? 
cc @pkudinov

idea404:
> This is a good point, thank you for bringing it up. For the near tracking enable/disable, will we auto-enroll them into enable or? cc @pkudinov

@TiffanyGYJ new users are unenrolled by default. Users that already have the NEAR CLI carry their settings with them from the previous version. 

We could also have users enrolled by default and adapt the notification. Both approaches are fine by me. We could also add to the message that account IDs are not coupled to the data, which could also help motivate data sharing. 

cc: @thisisjoshford @dgorton

volovyk-s:
I do not think that we should track users by default. Possible options are either ask and track/not track or do not track by default (I have described the downsides of this approach above).

PR #1019, opened 2022-08-29
added new clear state feature to clear contract state

doriancrutcher:
This feature utilizes Illia's new contract that gets deployed onto an account to have it's state cleared. 

for now this is only configured for testnet, would love for someone to take a look at this  and let me know your thoughts. then I'll add the option and configuration for mainnet


agileurbanite:
+1

mikedotexe:
Cool idea

Issue #949, opened 2022-01-08
generate-key with seed-phrase return wrong keypair


hanakannzashi:
***Describe the bug*** 

When running `near generate-key --seedPhrase="your seed phrase"` (without providing `--account_id <account>`) for multiple times, every time **a new random (rather than fixed) public key** is returned even if the **seed phrase keeps the same**. 

Example:

```bash
# 1st time
% near generate-key --seedPhrase="cow moon right send now cool dense quark pretty see light after"
Key pair with ed25519:9ETpmfmvHf4qcYz7UvCZgZhSZLin5PEAenRJ2ed2U1be public key for an account "7a50602085f66e7d2b963c9f8e7e09d5f5c2733dfdb0a4eb523caf9f4a74c061"

# 2nd time
% near generate-key --seedPhrase="cow moon right send now cool dense quark pretty see light after"
Key pair with ed25519:EXmwRyfmtWiHRp76YnDCdgsYyfqqbrdguQ8zciW4NfNh public key for an account "c90a2a98b6f1216ce44df6a2d7839d4790da9240774678cc88f2c6ca74fd429a"
```

Besides, it's quite weird that two json files are generated under the `~/.near-credentials/testnet` folder: one is the right json file, while the other is wrong. 

***Screenshots***
![1](https://user-images.githubusercontent.com/95295094/203999469-3e6e9f02-5861-44e7-943c-a15e205e7cc7.png)

***Update***
I fixed this issue on [PR#1036](https://github.com/near/near-cli/pull/1036)

think-in-universe:
@hanakannzashi hey, feel free to add more description in the issue as it's hard to understand what the problem is.

PR #1036, opened 2022-11-01
fix: generate-key-using-seed-phrase

hanakannzashi:
fix issue #949 

* The wrong keyStore variable was used before, resulting in two key pair files generated when using `near generate-key --seedPhrase="xxx"`


think-in-universe:
The fix make sense to me. @volovyks

volovyks:
Thank you @hanakannzashi ! It would be great to add a simple test for this change.
And in general, we need to turn on tests from external contributors in this repo. cc @ailisp I think it should be a standard for all repos where we expect contributions, I suppose, all tooling repos.

PR #1046, opened 2023-01-07
feat: Allow --force argument on deletion

mfornet:
Skip confirmation if --force is specified during deletion.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ const capture = require('./utils/capture-login-success');
 const inspectResponse = require('./utils/inspect-response');
 const eventtracking = require('./utils/eventtracking');
 const checkCredentials = require('./utils/check-credentials');
-const { askYesNoQuestion } = require('./utils/readline');
+const {askYesNoQuestion} = require('./utils/readline');
 
 // TODO: Fix promisified wrappers to handle error properly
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -233,8 +233,8 @@ exports.deleteAccount = async function (options) {
             throw e;
         }
     }
-
```
volovyks, line 236: Let's avoid formatting changes.

index.js
```diff
@@ -233,8 +233,8 @@ exports.deleteAccount = async function (options) {
             throw e;
         }
     }
-    
-    if (await confirmDelete()) {
+
+    if (options.force || await confirmDelete()) {
```
volovyks, line 237: It would be cool to add a note about this flag in README.
Small test is also appritiated.

mfornet: Added note on README. I tried to run integration tests locally, but doing `yarn test` is insufficient. I guess I need to spin off a NEAR node locally to make it work.

index.js
mfornet: Reverted

index.js
volovyks: It should work with testnet, but ok.

volovyks (approved):
Thnaks @mfornet !

PR #1057, opened 2023-03-20
Error message for undefined network includes network name passed

soonnear:
Updated error message for undefined network to include the name of the network that was passed

Previously, I was unsure if my commands were calling from 'localnet' or 'testnet'. I kept getting errors that it was an undefined network. Turns out, following the tutorials for localnet would pass in the name 'node0'. I spent hours debugging the wrong issue as I did pass in the correct network name. Hopefully, this more detailed error message would be helpful for beginners to near-cli in future.


volovyks (approved):
Nice fix @soonnear ! Thank you.

PR #1082, opened 2024-01-23
[v4.0.0] Reorganize, Reduce and Enhance 

gagdiez:
This is a proposal for a new **major** version of `near-cli`.

The main change is that the code was re-organized so it is simpler to maintain:
- All commands are now in the `commands` folder.
- Added multiple tests (all now running directly on testnet)
- All packages were updated (node v16 is now the minimum version, which is understandable)

The second change is that the complexity of `near-cli` was reduced by removing a couple of features:
- Removed analytics
- Breaking: removed `x-api` option.
- Breaking: removed `stake` command.
- Breaking: removed all `evm` commands.
- Breaking: Removed all `Ledger` support.
- Breaking: removed all `validator` commands.
- Breaking: removed `walletUrl`, `nodeUrl` and `helperUrl` options.
- Breaking: removed `dev-deploy` command. Users must now create an account (can fund it with a faucet) and deploy to it.

The third change is that Errors are now handled on each command, and not through some obscure middleware.

The fourth change is that a couple of fixes were implemented:
- `create-account` can now create `TLA`, `sub-accounts` and `.testnet/.near` accounts. The user can choose to fund the new account with an existing account or with a faucet service. 
- Fix: Default wallet is now `MyNearWallet` install of `NearWallet`
- Fix: Removed global flags, since they were not so global at all.. now each command has only the strict and necessary flags that it needs.
- Fix: Fixed `--networkId` flag, now it works as expected (i.e. switches network) in all relevant commands.
- Chore: many commands were renamed, but aliases are set in place for retro-compatibility. 

Fifth: besides the removal of commands explained in the second point, there is just one more breaking change: 
- The `generate-key` command does **not** save the key by default, now it needs the `--saveImplicit` flag to save the key. Also, it does not overwrite the key if it exists, it needs the `--force` flag for that.


gagdiez:
Comparison:

> All commands and options for `near-cli 4.0.0`
Screenshot 2024-01-23 at 13 25 15

> Commands and options in the same space for current `near-cli`
Screenshot 2024-01-23 at 13 26 41
Screenshot 2024-01-23 at 13 27 01

volovyks:
Lots of good changes. Nice work @gagdiez!
Are you sure we want to delete some of the features? Lise staking and ledger?
Overall we should move to the Rust version and sunset this one.

gagdiez:
@volovyks 

If we have `stake` we should implement `unstake` and `withdraw`. So I prefer people to use real wallets for it.

For Ledger, I saw it added a lot of complexity to the code... thought I might be evaluating the situation wrong here. Also, don't want a Ledger to be connected to a project with minimum support.

volovyks:
@gagdiez `stake` for users.
Ok, we may ask users to use Rust version for Ledger.

gagdiez:
@volovyks should we then implement `stake`, `unstake` and `withdraw`? I do not like the idea of having only `stake`, since then anyway you need a wallet to unstake it.

Otherwise, we can remove it now, and bring it back later if anyone was handling their staking through the CLI (which again, would be weird, since they cannot unstake it through there later)

gagdiez:
Merged since it is simpler to start fixing this version that keep working on the previous one (for example, TLA account creation is now broken)

PR #1099, opened 2024-02-26
feat: Add detailed instructions for installing near-cli

jancionear:
The installation guide in `README.md` provides a brief explanation of how to install the tool, but it isn't as detailed as it could be. For someone who isn't familiar with the javascript ecosystem the whole thing is pretty confusing - why `nvm`? Can I just do `apt install npm`? Why not? It took me some time before I found a sequence of instructions that worked.
[And users also run into problems](https://near.zulipchat.com/#narrow/stream/422293-pagoda.2Fcore.2Fstake-wars-iv/topic/questions.20and.20feedback/near/421267934)

It would be great to have a detailed step-by-step instruction on how to install `near-cli`. So this PR adds the exact list of steps needed to install `near-cli` on Ubuntu 20.04 LTS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,22 @@ _Click on a command for more information and examples._
 npm install -g near-cli
 ```
 
+For example, on Ubuntu 20.04 `near-cli` can be installed by running:
+```bash
+# Install nvm (https://github.com/nvm-sh/nvm?tab=readme-ov-file#installing-and-updating)
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
+source ~/.bashrc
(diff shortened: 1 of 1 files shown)


gagdiez:
hi @jancionear , thanks for the suggestion!

Do you think it is better to explain how to use `nvm` to install node or how to install node by [downloading the binaries](https://nodejs.org/en/download) and use them? The first option is easier, the second is more universal across distros

jancionear:
Another user just installed `near-cli` incorrectly: https://t.me/near_stake_wars/976

Could we merge those detailed instructions?

gagdiez (approved):

Issue #1107, opened 2024-03-03
New `NEAR_NETWORK` doesn't handle `NEAR_ENV` value


mikedotexe:
**Describe the bug**
Until a month ago, NEAR CLI has used `NEAR_ENV` when setting the network, like `mainnet` and `testnet`.
If folks have differing values on those two environment variables, it will use the old one that's not mentioned in the documentation: https://docs.near.org/tools/near-cli

<img width="1081" alt="Screenshot 2024-03-03 at 11 36 46 AM" src="https://github.com/near/near-cli/assets/1042667/eac85c0b-d26b-4a02-8a64-26cc121de678">

**To Reproduce**
See screenshot above

**Expected behavior**
As a user, if I follow the NEAR CLI documentation, I should be able to correctly target the network of choice and not have confusion.

Or perhaps this is a simple order of operations where the newly declared environment variable (`NEAR_NETWORK`) takes precedence over `NEAR_ENV`.

PR #1109, opened 2024-03-04
fix: display warning on duplicated env vars

gagdiez:
Address #1107 

The `near-cli` now shows a warning:

`Warning: NEAR_NETWORK and NEAR_ENV are both set! We will use NEAR_NETWORK (${process.env.NEAR_NETWORK})`


gagdiez:
addresses #1107

